PR #30, opened 2026-03-02
fix(provider-core): case-insensitive model validation

kotevcode:
## Summary

- `validateRequestModel` was doing a strict case-sensitive `includes()` check against `allowedModels`
- DHT topics are normalized to lowercase, so buyers discover and request models using lowercase names (e.g. `deepseek-r1`, `kimi-k2.5`)
- Providers configured with mixed-case `ANTSEED_ALLOWED_MODELS` (e.g. `deepseek-ai/DeepSeek-R1`) were returning 403 on every request

## Fix

Normalize both the incoming request model and the allowed list to lowercase before comparing.

## Test plan

- [ ] All existing `provider-core` tests pass (`pnpm --filter @antseed/provider-core test`)
- [ ] Buyer requesting `deepseek-r1` matches provider with `ANTSEED_ALLOWED_MODELS=deepseek-ai/DeepSeek-R1`
- [ ] Buyer requesting `KIMI-K2.5` matches provider with `ANTSEED_ALLOWED_MODELS=kimi-k2.5`

### DIFF
--- a/packages/provider-core/src/http-relay.test.ts
+++ b/packages/provider-core/src/http-relay.test.ts
@@ -348,6 +348,28 @@ describe('HttpRelay', () => {
     expect(parsed.model).toBe('moonshotai/kimi-k2.5');
   });
 
+  it('accepts model names case-insensitively', async () => {
+    fetchMock.mockResolvedValueOnce(new Response('{}', { status: 200 }));
+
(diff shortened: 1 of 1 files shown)


kotevcode:
@greptile

PR #100, opened 2026-03-13
fix(cli): preserve seller middleware and skills config

kotevcode:
## Summary
- preserve seller middleware entries when loading config.json
- preserve seller middleware confidentiality prompt and skillsDir
- add a regression test covering middleware and skills config loading

## Verification
- /Users/shahafan/Development/antseed/node_modules/.bin/tsx --test apps/cli/src/config/loader.test.ts
- verified earlier in the main workspace before isolating this branch; the temporary worktree does not have its own installed toolchain


kotevcode:
@greptile

PR #102, opened 2026-03-13
fix(provider): preserve final streaming with agent skills

kotevcode:
## Summary
- keep agent skill-loading iterations buffered but use a real final upstream stream for the buyer-visible response
- adapt buffered SSE bodies into Anthropic and Responses streaming formats when the seller returns a buffered stream body
- add regression tests for both the agent-provider streaming path and buffered SSE protocol adaptation

## Verification
- pnpm --filter @antseed/provider-core exec vitest run src/agent-provider.test.ts
- pnpm --filter @antseed/node run test -- service-api-adapter.test.ts
- pnpm run build:tier0 && pnpm run build:tier1 && pnpm --filter @antseed/cli run build

### DIFF
--- a/apps/cli/src/cli/commands/seed.ts
+++ b/apps/cli/src/cli/commands/seed.ts
@@ -1,8 +1,8 @@
 import type { Command } from 'commander'
 import chalk from 'chalk'
 import ora from 'ora'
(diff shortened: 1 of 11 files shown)


kotevcode:
@greptile

kotevcode:
@greptile

PR #104, opened 2026-03-14
feat: desktop protocol detection for OpenAI peers

kotevcode:
## Summary
- Desktop app now detects the peer's service API protocol (`anthropic-messages` vs `openai-chat-completions`) from the service catalog and uses the appropriate streaming implementation
- For OpenAI-protocol peers, uses pi-ai's built-in `streamOpenAICompletions` instead of the custom Anthropic SSE parser + buyer proxy format translation — more robust and maintained upstream
- For Anthropic-protocol peers, keeps the existing custom stream function (no change)
- Antseed routing headers (`x-antseed-provider`, `x-antseed-prefer-peer`) are set directly on the model's `headers` field

## Test plan
- [x] Desktop app builds successfully
- [ ] Test with Anthropic-protocol peer — verify existing behavior unchanged
- [ ] Test with OpenAI-protocol peer — verify streaming, tool calls, thinking work via pi-ai's native parser

### DIFF
--- a/apps/desktop/src/main/pi-chat-engine.ts
+++ b/apps/desktop/src/main/pi-chat-engine.ts
@@ -2230,23 +2230,24 @@ function createBuyerProxyOpenAIStreamFn(
 
         let sseBuffer = '';
         const decoder = new TextDecoder();
-        let currentBlock: AssistantMessage['content'][number] | null = null;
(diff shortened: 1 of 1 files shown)


kotevcode:
@greptile

PR #114, opened 2026-03-15
fix(bound-agent): wrap buyer system prompt as client context

kotevcode:
## Summary
- Instead of merging the buyer's system prompt with the agent's persona, the buyer's prompt is now wrapped in `<client-context>` tags and reframed as metadata about the user's environment
- The agent's persona remains authoritative — the model no longer picks up identity instructions from the buyer (e.g. "you are AntStation")
- The buyer's context is preserved so the agent can still reference it if useful

## Test plan
- [x] All 78 bound-agent tests pass
- [ ] Deploy to velvet-signal-social-media and verify "who are you" returns Velvet Signal, not AntStation


kotevcode:
@greptile

PR #162, opened 2026-03-22
feat(desktop): add browser preview panel and dev server tools

kotevcode:
## What

Add an embedded browser preview panel to the desktop app so users can see live web output alongside the chat.

- New `BrowserPreview` component with navigation toolbar, loading/empty states, and status bar
- `open_browser_preview` agent tool — opens a URL in the preview panel
- `start_dev_server` agent tool — launches a dev server as a background process, waits for readiness, returns the URL
- Split `ChatView` layout: chat on the left, preview panel on the right
- `webviewTag: true` enabled in the main window for embedded browsing
- Browser preview tools extracted into dedicated `browser-preview-tools.ts` module

## Why

When building web projects through the desktop agent, users had no way to see their work without switching to an external browser. The preview panel keeps the feedback loop inside the app.

## Testing

- Manual: started dev servers (Vite, Next.js) via the agent tool and verified preview renders correctly
- Build passes with `pnpm run build`

### DIFF
--- a/apps/desktop/src/renderer/ui/components/Sidebar.module.scss
+++ b/apps/desktop/src/renderer/ui/components/Sidebar.module.scss
@@ -224,22 +224,14 @@
 }
 
 .peer-group-convs {
-  display: grid;
-  grid-template-rows: 0fr;
-  transition: grid-template-rows 0.2s ease;
+  display: none;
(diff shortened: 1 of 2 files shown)


kotevcode:
@greptile

PR #309, opened 2026-04-14
fix(cli): respect --data-dir for daemon and buyer state files

kotevcode:
## Summary
- Several CLI commands hardcoded `~/.antseed` for `daemon.state.json` and `buyer.state.json`, so running with a custom `--data-dir` (e.g. in k8s deployments) made status commands show empty tables and broke buyer connection session overrides.
- `seller status` also never derived the wallet address from `ANTSEED_IDENTITY_HEX`, so the "become a provider" docs flow showed "not configured" even when the env var was set.
- `network browse` always did a fresh 30s DHT round-trip, even when a buyer daemon had already populated the peer cache.

## Changes
- `getNodeStatus` now takes `dataDir` and reads `<dataDir>/daemon.state.json`; `seller status` and `buyer status` pass `globalOpts.dataDir` through.
- `seller status` falls back to `loadCryptoContext(dataDir)` so the wallet address resolves from `ANTSEED_IDENTITY_HEX` or `identity.key` even before the daemon has started (matches `buyer status`).
- `getLocalSeederInfo` in `buyer start` takes `dataDir`.
- `BuyerProxy` takes `dataDir` in its config and stores the state-file path as an instance field; `buyer connection get/set/clear` and `buyer status` read/write `<dataDir>/buyer.state.json`.
- `network browse` reads discovered peers from `<dataDir>/buyer.state.json` (with fallback to `~/.antseed` for robustness) before starting an ephemeral node.

## Test plan
- [ ] \`pnpm --filter @antseed/cli run typecheck\` passes
- [ ] \`pnpm --filter @antseed/cli run test\` passes (4 pre-existing ChannelStore/metering failures are unrelated)
- [ ] \`antseed seller status\` with \`ANTSEED_IDENTITY_HEX\` set (no daemon running) shows the derived wallet address
- [ ] \`antseed seller status --data-dir <custom>\` against a seller daemon started with the same \`--data-dir\` shows populated metrics (peers, earnings, uptime)
- [ ] \`antseed network browse\` with a running buyer daemon returns instantly from the cached peer list

### DIFF
--- a/apps/cli/src/cli/commands/buyer/connection.ts
+++ b/apps/cli/src/cli/commands/buyer/connection.ts
@@ -1,11 +1,9 @@
 import type { Command } from 'commander'
 import { readFile, writeFile, rename } from 'node:fs/promises'
 import { join } from 'node:path'
-import { homedir } from 'node:os'
 import { randomUUID } from 'node:crypto'
 import chalk from 'chalk'
-
-const BUYER_STATE_FILE = join(homedir(), '.antseed', 'buyer.state.json')
+import { getGlobalOptions } from '../types.js'
 
 interface BuyerStateFile {
   state: 'connected' | 'stopped'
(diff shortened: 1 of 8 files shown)


kotevcode:
@greptile check the fixes

PR #318, opened 2026-04-16
feat(network-stats): add leaderboard with settlement indexing

kotevcode:
## Summary

- **Index `ChannelSettled` events** from the Channels contract alongside the existing `MetadataRecorded` indexer, storing per-event data with USDC amounts, token counts, and block timestamps
- **New `GET /leaderboard` endpoint** with query params: `role` (seller|buyer), `period` (day|month|all), `date` (YYYY-MM-DD), `limit` — returns ranked entries by USDC volume
- **Correlates settlement and metadata events** from the same transaction to capture both financial (delta, fees) and usage (tokens, requests) data in a single `settlement_events` table

### Changes by package

**`@antseed/node`** — `getChannelSettledEvents()` added to `ChannelsClient`, `DecodedChannelSettled` type exported

**`apps/network-stats`**:
- `store.ts` — new `settlement_events` table with timestamp indexes, `applySettlementBatch()`, `getLeaderboard()` with `periodBounds()` helper
- `settlement-indexer.ts` — new indexer that fetches `ChannelSettled` + `MetadataRecorded` events, resolves seller→agentId, stores atomically
- `server.ts` — new `GET /leaderboard` endpoint with validation and settlement indexer sync status
- `index.ts` — wires up `SettlementIndexer` when `channelsDeployBlock` is configured
- `store.test.ts` — 5 new tests (batch insert, day/month filtering, ranking, deduplication)

## Test plan

- [x] All 41 existing + new tests pass (`node --test dist/*.test.js`)
- [x] TypeScript strict mode passes for both `@antseed/node` and `apps/network-stats`
- [ ] Manual: deploy and verify `GET /leaderboard?role=seller&period=day` returns ranked results
- [ ] Manual: verify settlement indexer catches up from `channelsDeployBlock` on first run

### DIFF
--- a/apps/network-stats/src/index.ts
+++ b/apps/network-stats/src/index.ts
@@ -1,12 +1,13 @@
 import { mkdirSync } from 'node:fs';
 import { dirname } from 'node:path';
 
-import { StakingClient, StatsClient, resolveChainConfig } from '@antseed/node';
+import { ChannelsClient, StakingClient, StatsClient, resolveChainConfig } from '@antseed/node';
 
 import { NetworkPoller } from './poller.js';
 import { createServer } from './server.js';
 import { SqliteStore } from './store.js';
 import { MetadataIndexer } from './indexer.js';
+import { SettlementIndexer } from './settlement-indexer.js';
 
 const PORT = parseInt(process.env['PORT'] ?? '4000', 10);
 const CACHE_PATH = process.env['CACHE_PATH'];
@@ -25,6 +26,7 @@ const chainConfig = resolveChainConfig({
 });
 let store: SqliteStore | null = null;
 let indexer: MetadataIndexer | null = null;
+let settlementIndexer: SettlementIndexer | null = null;
 let stakingClient: StakingClient | null = null;
 
 if (chainConfig.statsContractAddress && typeof chainConfig.statsDeployBlock === 'number') {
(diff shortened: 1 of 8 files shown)


Augustas11:
Nice end-to-end pass. Tx-correlation between `ChannelSettled` and `MetadataRecorded` (via `metadataByTx.get(event.txHash).find(m => m.channelId === event.channelId)` at `store.ts:486-488`) is the cleanest way to merge financial + token-count data without needing a second indexer pass. The atomic `applySettlementBatch` transaction and the `INSERT OR IGNORE` on `(tx_hash, log_index)` cover the re-index + retry safety story.

A few real issues to address before merge:

1. **Precision loss in the leaderboard SUM** — `store.ts:756, 770, 786, 802` all use `SUM(CAST(delta_amount AS INTEGER))`. SQLite's INTEGER is 64-bit, but `better-sqlite3` returns SUM results as a JavaScript Number unless `safeIntegers` is configured (it isn't in `store.ts`). JS Number tops out at 2^53 ≈ $9B for USDC base units. The returned `totalRevenue: String(r.total_revenue)` only stringifies after the precision is already lost at the SQL→JS boundary. Two ways to fix:
   - `db.defaultSafeIntegers(true)` (or per-statement `.safeIntegers()`) → returns BigInt
   - Sum in JS: load events for the period and `total += BigInt(row.delta_amount)`
   For current AntSeed volume the bug is theoretical, but it's invisible when it bites and the fix is one line.

2. **`agent_id = 0` sinks all unstaked traffic into one leaderboard row.** Currently `applySettlementBatch` writes `agentIdBySeller.get(event.seller) ?? 0` (`store.ts:480`), and `getLeaderboard('seller', ...)` groups by `agent_id` without filtering. If unstaked sellers settle (e.g., during a stake-revoke window), their volume all aggregates under `id: '0'` and may top the leaderboard. Either add `WHERE agent_id > 0` to the seller queries or skip the insert when `agentId === 0`.

3. **No tick-level test.** The orchestration in `settlement-indexer.ts:81-168` (chunking, parallel agent-id resolution, parallel block-timestamp fetch, atomic apply) has all the moving parts. The five new store tests cover the storage layer but not the integration. A test stubbing `channelsClient.getChannelSettledEvents` + `statsClient.getMetadataRecordedEvents` + `provider.getBlock` would lock in the contract.

4. **Deep reorg cleanup.** `INSERT OR IGNORE` is correct for re-index safety, but if a settled tx gets orphaned by a >12-block reorg, the row stays in the leaderboard forever. Not blocking — but worth a TODO and an alert path on `getCheckpoint > head - reorgSafetyBlocks * 2`.

5. **Silent failure when metadata correlation misses.** When `metadataByTx.get(event.txHash)?.find(...)` returns nothing (e.g., AntseedStats write reverted in a try/catch — see INV-L3), token counts default to 0 with no log. A `console.warn(`no metadata for settle ${event.channelId.slice(0,12)} in tx ${event.txHash.slice(0,12)}`)` would surface H5-class bugs.

6. **Minor: `m.channelId.toLowerCase() === event.channelId`** at `store.ts:488` — `event.channelId` is already lowercased on decode (`channels-client.ts:200`), but `m.channelId` is `.toLowerCase()`'d defensively. Worth confirming `DecodedMetadataRecorded.channelId` is also lowercased on its own decode for symmetry.

INV-34 / INV-62 / H5 not affected — pure read indexer.

PR #322, opened 2026-04-18
fix(node): seller validates requested service against published list

kotevcode:
## Summary
- Seller now rejects requests whose `service`/`model` does not match one of the provider's published services, instead of blindly forwarding to upstream.
- Previously a shortcut in `matchProvider` (`providers.length === 1`) caused single-provider sellers — the most common deployment — to match anything, including requests with no `model` field at all. Upstream then rejected with opaque errors like "you must choose a model", and payment reserves were wasted on requests we were never going to serve.
- `_extractRequestedService` now reads `service ?? model` to match the buyer-side and provider-relay conventions. Provider relays already strip `service` and normalize to `model` before the upstream fetch (`packages/provider-core/src/http-relay.ts`, `plugins/provider-openai-responses/src/index.ts`), so upstreams still only see `model`.
- Missing/unknown service returns `400` with `invalid_request_error` (`model_required` or `model_not_found`) before any payment handshake, 402, or upstream dispatch. The old `502 "No matching provider"` fallback is removed — the hoisted check covers it.

## Test plan
- [x] `pnpm --filter=@antseed/node run typecheck` passes.
- [x] `pnpm --filter=@antseed/node test` — all 537 tests pass (including the existing `matchProvider` test in `node-payment-pricing.test.ts`).
- [ ] Manual: run a seller with a single provider, POST to `/v1/chat/completions` with no `model` — expect `400 model_required` locally, no upstream call, no payment reserve.
- [ ] Manual: POST with `model` that is not in `provider.services` — expect `400 model_not_found`.
- [ ] Manual: POST with `service: "..."` instead of `model` and a matching published service — request succeeds end-to-end.

### DIFF
--- a/packages/node/src/seller-request-handler.ts
+++ b/packages/node/src/seller-request-handler.ts
@@ -65,14 +65,38 @@ export class SellerRequestHandler {
         return;
       }
 
+      // Match the requested model to one of our published services BEFORE any
+      // payment handshake or upstream forwarding. Rejecting unknown/missing
+      // models locally avoids reserving payment for something we won't serve
+      // and surfaces a clear error instead of an opaque upstream 4xx.
+      const provider = this.matchProvider(request);
+      if (!provider) {
+        const requestedService = this._extractRequestedService(request);
+        const allServices = this._deps.providers.flatMap((p) => p.services);
+        const errorMessage = requestedService === null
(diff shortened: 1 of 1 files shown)


kotevcode:
@claude review here

PR #324, opened 2026-04-19
fix(node,cli): stabilise peer discovery against scan jitter

kotevcode:
## Summary

Peers were disappearing from the buyer when a single DHT scan happened to miss them. Two cliffs lined up:

- Buyer-side `CARRY_FORWARD_TTL_MS` (30 min) equalled `PeerLookup.maxAnnouncementAgeMs` (30 min), so one unlucky scan evicted the peer until the next successful cycle.
- Announce-side failures in `PeerAnnouncer._tryAnnounceTopic` were swallowed silently and waited the full 15-min reannounce interval before retrying, leaving the seller off the DHT long enough to hit the buyer staleness cutoff.

We verified the failure mode live: `The Seeder` was absent from `buyer.state.json` at one point but three back-to-back cold DHT lookups found it, and its `/metadata` endpoint was healthy the whole time — classic scan jitter colliding with a tight carry-forward window.

### Changes

**Buyer cache decouples from DHT staleness** — `apps/cli/src/proxy/buyer-proxy.ts`, `packages/node/src/types/peer.ts`
- `CARRY_FORWARD_TTL_MS`: 30 min → 2 h. Still bounded, but no longer cliff-aligned with the 30-min staleness cutoff.
- New optional `PeerInfo.lastReachedAt`, stamped in `_rememberSuccessfulPeer` on every successful transport contact and persisted to `buyer.state.json`.
- `parsePersistedPeers` and `_replacePeers` anchor freshness on `max(lastSeen, lastReachedAt)`, so a peer we've actually talked to survives transient DHT staleness even when its announcement record ages out.
- `_replacePeers` preserves the prior `lastReachedAt` when a peer is re-observed from DHT (the announce doesn't carry it).

**Announce failures stop being silent** — `packages/node/src/discovery/announcer.ts`
- `_tryAnnounceTopic` now returns `boolean`, logs failures via `debugWarn`, and optionally calls `DHTHealthMonitor.recordAnnounce` (new `healthMonitor?` field on `AnnouncerConfig`).
- `_announceTopics` aggregates per-cycle failures.
- New `_scheduleRetryAfterFailure` schedules an exponential-backoff retry (60s → 120s → 5m → 10m, capped at `reannounceIntervalMs`) when any topic fails; success resets the backoff; `stopPeriodicAnnounce` clears any pending retry.

## Test plan

- [x] `pnpm --filter @antseed/node typecheck` — clean
- [x] `pnpm --filter @antseed/cli typecheck` — clean (after node rebuild)
- [x] `pnpm --filter @antseed/node test` — 537/537 pass
- [x] `pnpm --filter @antseed/cli test` — 64/64 pass (incl. updated `parsePersistedPeers` staleness test for the new 2h window)
- [x] `apps/desktop` main `tsc -p tsconfig.main.json` — clean
- [ ] Manual: run a buyer for >1 h, confirm peers that flicker out of individual DHT scans no longer disappear from `buyer.state.json`

## Follow-ups (not in this PR)

- Wire a `DHTHealthMonitor` instance in `node.ts` so `recordAnnounce` is actually consumed, and trigger a DHT re-bootstrap when the routing table collapses.
- Tighten `HttpMetadataResolver` cooldown — the current 30s → 30min exponential blackout can silently suppress a recovered peer for up to 30 min per lookup.
- Consider a cheap liveness probe on carry-forward peers before marking them `online` in the dashboard (today `online` is derived from `lastSeen` alone).

### DIFF
--- a/apps/cli/src/proxy/buyer-proxy.test.ts
+++ b/apps/cli/src/proxy/buyer-proxy.test.ts
@@ -96,7 +96,7 @@ test('selectCandidatePeersForRouting can still include peers without service pro
 // blocking on DHT discovery.
 
 const validPeerId = 'a'.repeat(40)
-const MAX_AGE_MS = 30 * 60_000
+const MAX_AGE_MS = 2 * 60 * 60_000
 const NOW = 1_700_000_000_000
 
 test('parsePersistedPeers returns [] for null/undefined/junk input', () => {
--- a/apps/cli/src/proxy/buyer-proxy.test.ts
+++ b/apps/cli/src/proxy/buyer-proxy.test.ts
@@ -139,13 +139,68 @@ test('parsePersistedPeers drops entries with non-array providers', () => {
   assert.equal(result.length, 0)
 })
 
-test('parsePersistedPeers drops entries with stale or missing lastSeen', () => {
+test('parsePersistedPeers drops entries with stale or missing freshness anchors', () => {
   const result = parsePersistedPeers(
     {
       discoveredPeers: [
         { peerId: validPeerId, providers: ['openai'], lastSeen: NOW - MAX_AGE_MS },
         { peerId: 'b'.repeat(40), providers: ['openai'] },
         { peerId: 'c'.repeat(40), providers: ['openai'], lastSeen: 'nope' },
+      ],
+    },
+    NOW,
+  )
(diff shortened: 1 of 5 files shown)


kotevcode:
@claude review this pr

PR #330, opened 2026-04-19
fix(desktop): allow parallel chat runs

SheferKagan:
## What

Allows desktop chat runs to continue independently per conversation instead of treating the whole chat UI as having one global in-flight request.

This keeps streaming/sending state tied to the conversation that started the run, and lets users switch chats or start another chat without incorrectly blocking unrelated conversations.

## Why

Parallel chat work is needed so one active stream does not freeze the whole desktop chat experience. Without this, a long-running response can make other conversations feel stuck or unavailable, and stream completion/error events can clear state for the wrong active chat after the user switches conversations.

## Testing

Tested as part of the combined desktop chat integration branch.

Validated with:

- `pnpm --filter=@antseed/desktop test`
- `pnpm --filter=@antseed/desktop run typecheck:renderer`
- `pnpm --filter=@antseed/desktop run build`

No dedicated automated test was added for this PR yet; coverage is currently through TypeScript/build validation and the existing desktop main-process tests.

## Checklist

- [x] `pnpm run build` passes
- [x] `pnpm run test` passes


kotevcode:
@claude

kotevcode:
Merged main and addressed the review feedback in 26cb397:

**Conflicts resolved**
- `apps/desktop/src/renderer/modules/chat.ts` and `apps/desktop/src/renderer/types/bridge.ts`: kept per-conversation sending state from this branch while preserving the new `stopReason` plumbing and richer error surfacing introduced on `main` (e.g. `stopReasonSummary`, `result.stopReason?.message`, unified `chatAiSendStream` return type).
- `chatAiAbort` signature now `(conversationId?: string)` — backwards compatible and used by the abort-path and stuck-request retry with the current conversation.

**Review items**
1. `deleteConversation` now also clears `sendingConversationIds`, `streamTurnsByConversation`, and `streamStartedAtByConversation` before `startNewChat()` (or calls `syncActiveConversationSendingState()` when deleting a non-active conversation). Prevents stale streaming indicator state after deleting an in-flight chat.
2. `setChatSending` fallback path (no active conversation) now calls `ensureStreamingIndicatorTimer()` when `sending` is true so the elapsed timer ticks during the pre-conversation window.
3. `syncActiveConversationSendingState` collapsed the two mutually-exclusive `if (sending)` / `if (!sending)` blocks into a single `if/else`.

**Validation**
- `pnpm --filter=@antseed/desktop run typecheck:renderer` ✓
- `pnpm --filter=@antseed/desktop test` ✓ (30/30)
- `pnpm --filter=@antseed/desktop run build` ✓

Issue #124, opened 2026-03-16
Unable to upload HTML files - file upload functionality blocked


Amosme:
## Problem
It's impossible to upload HTML files through the AntSeed interface.

## Expected Behavior
HTML files should be uploadable like other file types.

## Actual Behavior
HTML file uploads are blocked or fail.

## Environment
- Platform: AntStation desktop client
- Date: March 16, 2026

## Impact
This prevents users from:
- Uploading website files
- Sharing HTML documentation
- Working with web development projects

## Suggested Fix
Review file upload restrictions and ensure HTML files are properly supported while maintaining security.

Reported via AntStation desktop client.

PR #334, opened 2026-04-20
feat(desktop): broad safe chat attachment support (#124)

SheferKagan:
Adds broad safe attachment handling for AntStation chat (HTML/PDF/Office/text/ZIP/images)
Keeps Pi native image path; non-images converted to safe text context
Adds limits/guardrails and ZIP safety checks
Updates renderer/preload/bridge for multi-attachment flow with file chips/cards
Adds attachment preparation tests
Verified:
pnpm --filter=@antseed/desktop run test
pnpm --filter=@antseed/desktop run typecheck:renderer
pnpm --filter=@antseed/desktop run build


kotevcode:
@claude review this pr

kotevcode:
**Review: feat(desktop): broad safe chat attachment support (#124)**

*kimi 2.6 running through antstation*

**Architecture Verdict:** This is the right approach for the current Pi-AI stack. Since `@mariozechner/pi-ai` only supports text + images natively, converting documents/archives to extracted text and wrapping them in XML-like blocks is pragmatic. Keeping images as first-class `ImageContent` attachments and running heavy parsing (`officeparser`, `fflate`) in the main process is architecturally sound. The IPC boundary receives sanitized `PreparedChatAttachment` objects, which is clean.

**Will it work?** Yes, for the supported file types (HTML, PDF, DOCX, text, ZIP, images). The `extractOfficeText` integration and ZIP in-memory extraction are correct. The prompt assembly and `session.prompt` wiring is consistent with Pi's multimodal API. However, there are a few issues that should be addressed before merging.

---

**Issues**

*High / Medium*

1. **Incomplete prompt-injection guard in `neutralizeFileTags`**
   `chat-attachments.ts:109` only escapes `&lt;/file&gt;`, but it does **not** escape `&lt;file`. If an attachment (text, ZIP entry, HTML, etc.) contains a literal `&lt;file name=&quot;...&quot;&gt;` followed by `&lt;/file&gt;`, the `convertPersistedAttachmentPromptToBlocks` regex in `pi-chat-engine.ts` will prematurely close at the *first* `&lt;/file&gt;` it encounters. This corrupts the file block when the conversation is reloaded from Pi history. The `safeAttribute` helper also doesn't touch the body content.
   **Fix:** also neutralize `&lt;file\b` in `neutralizeFileTags`, e.g.:
   ```ts
   function neutralizeFileTags(text: string): string {
     return text
       .replace(/&lt;file\b/gi, '&lt;\\file')
       .replace(/&lt;\/file&gt;/gi, '&lt;\\/file&gt;');
   }
   ```
   (and similarly `&lt;/zip-entry&gt;` if you ever parse those separately).

2. **Per-file size gate trusts renderer-reported `raw.size`**
   `prepareChatAttachments` (`chat-attachments.ts`) rejects based on `raw.size` *before* decoding. The actual decoded buffer size is `buffer.byteLength`. A manipulated IPC payload could claim `size: 0` while carrying a large base64 body, slipping past the per-file limit. The total-bytes gate catches it, but the per-file gate should be hardened.
   **Fix:** after `buffer = decodeBase64(raw)`, validate `buffer.byteLength &gt; limits.maxFileBytes` instead of (or in addition to) `raw.size`.

*Medium / Low*

3. **Missing `accept` attribute on the hidden `&lt;input type=&quot;file&quot;&gt;`**
   `ChatView.tsx` removed the `accept` attribute when switching from image-only to multi-file. Users can now select executables or binaries that will only be rejected after the main-process round-trip. Adding an `accept` filter improves UX and avoids unnecessary IPC overhead.

4. **Out-of-scope change in `process-manager.test.ts`**
   The path-normalization fix in `process-manager.test.ts` (using `join`/`resolve`) is a useful Windows fix, but it is completely unrelated to issue #124 / chat attachments. It should be split into its own PR.

*Informational*

5. **Error attachments vanish on conversation reload**
   Error attachments are filtered out of the prompt text by `buildAttachmentPromptText`. This means they are not stored in Pi history. When a conversation is reopened and rebuilt via `buildConversationFromManager` → `convertPiMessagesToUi`, those error file blocks disappear. This is an acceptable trade-off for now, but worth noting: the renderer's local storage is the only place that preserves the full `PreparedChatAttachment` metadata (including `status: 'error'`).

6. **Electron packaging risk with `officeparser` + `pdfjs-dist`**
   `pdfjs-dist@5.6.205` may attempt to require `canvas` (a native module) when running in Node.js. If `officeparser` triggers that path, the packaged Electron build will need a native rebuild for the correct Node ABI. Please verify that the desktop `build` script still works after these deps are installed, and that `require.resolve('pdfjs-dist/build/pdf.worker.mjs')` resolves correctly inside the packaged app.

7. **Large IPC payloads**
   At the allowed max (5 × 25 MiB), the base64 payload across IPC can reach ~165 MiB. Electron handles it, but it blocks the renderer briefly. This is a reasonable scope trade-off for a desktop app.

---

**What's done well**
- ZIP bomb defense: `inflatedBytes` + entry-count limits with clear skip messages.
- Path traversal guard: `isUnsafeZipEntryName` correctly catches `..`, absolute paths, and Windows drive letters.
- Error isolation: each attachment can fail independently; the batch never aborts.
- Native image path preserved: Pi's `ImageContent` is used directly, which is correct.
- Retry-after-payment works: `PreparedChatAttachment` objects are stashed on the `file` block in local storage and re-extracted cleanly in `retryAfterPayment`.

---

**Summary**
Ship it after fixing the `&lt;file` neutralization and the per-file size hardening. Those are the two real blockers. The architecture is solid and this is the right way to support attachments given the current Pi-AI multimodal constraints.

Issue #333, opened 2026-04-20
Buyer shows generic timeout when seller doesn't confirm payment lock — error should surface likely causes (e.g. low deposits)


phirex:
## Summary
When the buyer's deposits balance is too low for a seller's pricing/reserve requirements, the seller appears to accept the initial handshake but never sends a Lock Confirmation. The buyer then emits only:

```
502 P2P request failed: Lock confirmation timed out for seller <peerId>... (30000ms)
```

This error is indistinguishable from a seller that is genuinely offline, a seller that is overloaded, or a transient network hiccup — so integrators spend hours diagnosing the wrong thing.

## Reproduction
1. `antseed buyer deposit 1` — deposit a small amount (e.g. just under the seller's `reserveFloor`).
2. Pin to any seller whose reserve is higher than available deposits:
   `antseed buyer connection set --peer <peerId>`
3. Send any chat completion request through the proxy.
4. Observe the 30s timeout and the message above. No hint anywhere that low deposits could be the cause.

## What I expected
The buyer already knows:
- Current `deposits.available` from the state file.
- The seller's pricing and `reserveFloor` from `providerPricing.defaults` / browse metadata.
- That the seller responded `PaymentRequired` before going silent.

So when the lock-confirmation timeout fires, the buyer could produce:

```
502 Seller <peerId> did not confirm payment lock within 30s.
Likely causes:
  - Deposits available (0.87 USDC) are below seller reserveFloor (likely >= 10 USDC
    based on last PaymentRequired). Try: antseed buyer deposit <amount>.
  - Seller is overloaded or gated on subscription (try: antseed buyer subscribe status).
  - Seller is online but the payment process stalled (retry, or pin a different peer).
```

Even without the live arithmetic, just listing the three likely causes in the error body would save hours.

## Buyer log excerpt (real data, 4 peers tested in one session)
```
[PaymentMux] -> send SpendingAuth (755b)
[PaymentMux] <- recv PaymentRequired (195b)
[BuyerNegotiator] PaymentRequired from 4668854ba3e8... buffered
... 30s later ...
[proxy] Request failed after 30555ms: Lock confirmation timed out for seller 4668854ba3e8... (30000ms)
```
Same pattern observed against 3 different sellers (SecretAI, Open Forge, Dark Signal). Resolved immediately after topping deposits from 0.87 → 10.87 USDC.

## Suggested fixes
1. Include deposits vs. last-seen seller pricing in the timeout error body.
[...]

PR #337, opened 2026-04-21
fix(node): diagnose buyer lock-confirmation failures

kotevcode:
## Summary

The buyer's payment negotiation used to swallow the real cause of a failed payment lock behind a generic one-line error:

```
502 P2P request failed: Lock confirmation timed out for seller <peerId>... (30000ms)
```

This was indistinguishable from a seller that was offline, overloaded, or hitting a transient network hiccup, and operators burned hours looking in the wrong place (see #333).

This PR enriches the lock-failure error with a `Likely causes` block assembled from data the buyer already has locally (the reserve ceiling it authorized, the seller's `minBudgetPerRequest`) so operators get an actionable explanation without any extra RPC round-trips on the buyer side.

## Why zero buyer-side RPC calls

An earlier revision of this PR pre-checked the buyer's deposit balance before signing, and re-fetched it on every failure for the hint. That added up to 3 `getBuyerBalance` calls per negotiation on the buyer's endpoint, which is typically less reliable than the seller's. Reviewer feedback (thanks!) flagged this correctly.

The right place for that data is the seller:

- The seller is the one making the on-chain `reserve()` / `topUp()` / `settle()` calls.
- The seller has a reliable RPC endpoint (required for normal operation).
- The seller knows the exact revert reason and can fetch a fresh deposits snapshot cheaply.

So this PR's scope is narrowed to the buyer-local error-message enrichment. The protocol change that makes the seller send a structured `AuthNack` frame is tracked as **#341** and will plug into the error throw paths introduced here.

## Technical details

- `_waitForLockConfirmation` now takes a `{ requestedReserve, minBudgetPerRequest }` context and synthesizes the `Likely causes` block on timeout / rejection.
- Context is threaded into every call site: `_doNegotiatePayment`, `applyExternalSpendingAuth`, and both `_recoverExistingSession` paths (reserve replay + extend/resend current auth).
- No new RPC calls. The pre-existing `handle402` "available <= 0" credit gate still runs once up front, exactly as before.

## Example output

**Before:**
```
Lock confirmation timed out for seller 4668854ba3e8... (30000ms)
```

**After:**
```
Lock confirmation timed out for seller 4668854ba3e8... (30000ms).
Likely causes:
  - Deposits may be below the reserve ceiling the seller will lock (10 USDC).
    The seller's on-chain reserve() reverts with InsufficientBalance before AuthAck.
    Check with "antseed buyer status"; top up with "antseed buyer deposit <amount>".
  - Seller may be overloaded or gated on subscription (try another peer).
  - Seller may be online but the payment process stalled (retry, or pin a different peer).
```

Once #341 lands, the first bullet will be replaced by the seller's exact revert reason + a deposits snapshot the buyer didn't have to fetch.

## Test plan

`packages/node`: `pnpm test` \u2014 54 files / 561 tests pass. Three focused tests in `describe('lock failure diagnostics (issue #333)')`:

- `Lock rejected` errors include the reserve-ceiling hint and generic causes
- `Lock confirmation timed out` errors include the same hint block (verified with `vi.useFakeTimers`)
- The enriched hint never triggers additional buyer-side `getBuyerBalance` calls

`pnpm run build` \u2014 full workspace build passes.

Refs #333. Follow-up: #341.

### DIFF
--- a/packages/node/src/payments/buyer-payment-negotiator.ts
+++ b/packages/node/src/payments/buyer-payment-negotiator.ts
@@ -15,6 +15,7 @@ import type { SellerAddressResolver } from '../discovery/seller-address-resolver
 import { SellerAuthorizationError } from '../discovery/seller-address-resolver.js';
 import { parseResponseUsage } from '../utils/response-usage.js';
 import { computeCostUsdc, type ServicePricing } from './pricing.js';
+import { formatUsdc } from './usdc-utils.js';
 
 export interface BuyerNegotiatorConfig {}
 
@@ -471,7 +472,15 @@ export class BuyerPaymentNegotiator {
     pmux.sendSpendingAuth(payload);
     debugLog(`[BuyerNegotiator] External SpendingAuth sent to seller ${peer.peerId.slice(0, 12)}..., waiting for AuthAck...`);
 
-    await this._waitForLockConfirmation(peer.peerId);
+    let reserveCtxAmount: bigint | undefined;
+    try {
+      reserveCtxAmount = payload.reserveMaxAmount != null
+        ? BigInt(payload.reserveMaxAmount)
+        : BigInt(payload.cumulativeAmount);
+    } catch {
+      reserveCtxAmount = undefined;
+    }
+    await this._waitForLockConfirmation(peer.peerId, { requestedReserve: reserveCtxAmount });
     debugLog(`[BuyerNegotiator] AuthAck received from seller ${peer.peerId.slice(0, 12)}...`);
     this._lockedPeers.add(peer.peerId);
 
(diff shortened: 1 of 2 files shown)


kotevcode:
@claude review pls

kotevcode:
Thanks for the review! Acting on the #1 finding (up to 3 `getBuyerBalance` RPC calls per negotiation):

- Removed `_assertDepositsSufficient` entirely.
- Removed the `getBuyerBalance` call from `_buildLockFailureHint`.
- The hint is now assembled purely from data the buyer already has locally (`requestedReserve`, `minBudgetPerRequest`). No extra RPC load on the buyer endpoint.

The architectural argument is that the seller should drive these diagnostics \u2014 it has the reliable RPC endpoint and the exact revert reason. Tracked as #341: a new `AuthNack` frame on `PaymentMux` carrying the revert selector + buyer deposits snapshot, which the buyer will render via the error-throw paths set up in this PR. That change also connects the currently-dormant `markRejected()` code path and unblocks the buyer immediately instead of waiting 30s.

Tests trimmed accordingly \u2014 new test `does not call the buyer RPC during or after a failed lock confirmation` pins the zero-RPC guarantee.

Notes on the other review points:

2. **Timeout guard on `getBuyerBalance`** \u2014 now moot, no call in the hint path.
3. **`payment:required` fires before pre-check** \u2014 pre-check removed, so `payment:required` is only immediately followed by `payment:signed` or by the enriched lock error (which doesn't emit anything new).
4. **Recovery paths lacking `requestedReserve`** \u2014 acknowledged as a limitation in the code comment; the seller-side nack in #341 will supply that context directly.
5. **`applyExternalSpendingAuth` test coverage** \u2014 will fold a negative test into #341 where the full AuthNack-driven flow is exercised end-to-end.

PR #351, opened 2026-04-23
fix(node): Catch up buyer SpendingAuth on budget-exhausted 402

kotevcode:
## Symptom

Observed in production on the Open Forge (deepinfra) peer: the desktop client surfaces a raw `402` with the seller's JSON body after a few successful requests on an existing channel:

```
[proxy] Response: 402 (2332ms, 434 bytes)
[proxy] Upstream error detail: {"error":"payment_required","minBudgetPerRequest":"10000","suggestedAmount":"1000000","inputUsdPerMillion":0.36,...}
```

Seller log for the same request:
```
[SellerHandler] Budget exhausted for 232292ef... (spent=85119 >= accepted=56218)
  — returning 402, awaiting higher SpendingAuth / renegotiation
```

## Root cause

When the seller has `spent > accepted` (a prior request's cost blew past the pre-fund buffer, or a top-up stalled behind the per-buyer mutex), it sends `402 + PaymentRequired`. The buyer's `handle402` tried to recover via `extendCurrentSpendingAuth(peerId, minBudgetPerRequest=10_000)`, which advances cumulative by only `10_000` at a time. With `spent=85_119` and `accepted=56_218`, the gap (28_901) could never be closed in one step — every signed auth was rejected by the seller as underfunded, and eventually the raw seller 402 body leaked to the user.

The 402 payload did not include enough information for the buyer to close the gap.

## Fix

**Seller** (`seller-request-handler.ts`): the budget-exhausted branch now embeds `requiredCumulativeAmount` / `currentSpent` / `currentAcceptedCumulative` / `channelId` in both the 402 JSON body and the `PaymentRequired` frame, and also emits a `NeedAuth` on the PaymentMux so the buyer's standard auto-sign path can catch up without the 402 ever round-tripping to the user.

**Buyer — `buyer-payment-manager.ts`**: `extendCurrentSpendingAuth` accepts an optional absolute `targetCumulative`. When the seller reports a target, `verifiedCost` is advanced to it (same reasoning as the existing collapsed-overdraft reopener — the seller can already claim up to `accepted` on-chain, so the exposure is bounded) and the signed cumulative jumps straight to the seller's spend in one hop.

**Buyer — `buyer-payment-negotiator.ts`**: `handle402` parses `requiredCumulativeAmount` from the seller body and forwards it through `_recoverExistingSession` to `extendCurrentSpendingAuth`. The `returnPaymentRequired` path no longer passes the raw seller response back to the caller — it writes a buyer-authored body with a user-friendly "top up your deposits" message so internal catch-up / accounting / per-service pricing fields never leak.

**Protocol** (`types/protocol.ts`, `payment-codec.ts`): `PaymentRequiredPayload` gains optional `requiredCumulativeAmount`, `currentSpent`, `currentAcceptedCumulative`, `channelId`, and the previously undecoded `cachedInputUsdPerMillion`. All fields are optional — wire-compatible with older peers.

## Tests

Three new tests covering the regression:

- `payment-codec.test.ts`: PaymentRequired round-trips all new catch-up fields.
- `buyer-payment-manager.test.ts`: `extendCurrentSpendingAuth` reproduces the 56_218 → 85_119 Open Forge gap with `maxPerRequestUsdc=10_000` and asserts the buyer jumps to the target in one step.
- `buyer-payment-negotiator.test.ts`: `handle402` forwards `requiredCumulativeAmount` from the 402 body as the catch-up target, plus an updated insufficient-credits test asserting the returned body is buyer-authored (no raw seller catch-up/pricing fields).

All 564 `@antseed/node` tests pass. `pnpm run typecheck` and `pnpm run build` are green.

### DIFF
--- a/apps/cli/src/cli/commands/buyer/start.ts
+++ b/apps/cli/src/cli/commands/buyer/start.ts
@@ -299,7 +299,13 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
             ? String(Math.round(parseFloat(cryptoOverrides.defaultLockAmountUSDC) * 1_000_000))
             : '1000000',
           platformFeeRate: config.payments?.platformFeeRate,
-          maxPerRequestUsdc: config.payments?.maxPerRequestUsdc ?? '100000',
+          // $0.30 overdraft window per channel — large enough that a single
+          // typical long-context request (~$0.05–$0.15 on the priciest
+          // published models) fits within verifiedCost + maxPerRequest, so the
+          // budget-exhausted 402 catch-up closes in a single signature. Set
+          // conservatively to bound the worst-case exposure a malicious
+          // seller can extract via an inflated 402 target (per 402 round trip).
+          maxPerRequestUsdc: config.payments?.maxPerRequestUsdc ?? '300000',
           maxReserveAmountUsdc: config.payments?.maxReserveAmountUsdc ?? '1000000',
         }
       }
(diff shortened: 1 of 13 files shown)


kotevcode:
@claude review this

PR #352, opened 2026-04-23
fix(buyer): keep chat pinned to its peer/service

kotevcode:
## Description

A single 300s stream timeout was evicting the failing peer from the buyer proxy cache, which dropped its `providerServiceApiProtocols` metadata. On the next DHT refresh the peer sometimes re-hydrated without the chat's service listed, and the desktop renderer's `applyChatServiceOptions` then silently rebound the active chat to an unrelated peer's service via the first-in-catalog fallback. Outbound requests went out with `pin-peer=<originalPeer>` + `service=<other peer's service>` and stuck in a permanent "Service strict-miss" 502 retry loop.

Two changes, one on each side of the broken flow:

- **Buyer proxy (`apps/cli/src/proxy/buyer-proxy.ts`)** — replace single-failure eviction with a rolling `_peerFailures` counter (threshold 4, window 5 min). Counters clear on success and on actual eviction. The existing short-circuits for `/v1/models` probes and connection-churn-with-healthy-replacement are preserved, now also flowing through the softened counter.
- **Desktop renderer (`apps/desktop/src/renderer/modules/chat.ts`)** — drop the `optionCandidates[0]?.value` first-option fallback in `applyChatServiceOptions` when a conversation is active. If neither the current selection nor the conversation's persisted `(service, provider, peerId)` match the refreshed catalog, leave `chatSelectedServiceValue` unchanged so the chat stays bound to its original peer/service. New-chat start-screen behavior is unchanged.

## Release Notes

- Fixed a case where a slow/stuck chat would cause the desktop to silently switch to a different peer's service, leaving the chat stranded on repeated "No peers support …" errors. The original peer now stays in cache through a single timeout, and active chats stay pinned to their chosen service.

## Types of Changes

## Checklist

- [ ] I have added tests (if appropriate)

### DIFF
--- a/apps/cli/src/proxy/buyer-proxy.ts
+++ b/apps/cli/src/proxy/buyer-proxy.ts
@@ -93,6 +93,16 @@ const RETRYABLE_STATUS_CODES = new Set([408, 429, 500, 502, 503, 504])
  */
 const CARRY_FORWARD_TTL_MS = 2 * 60 * 60_000
 
+/**
+ * A single failed request is not enough to evict a peer: timeouts, transient
+ * upstream slowness, or one flaky stream shouldn't wipe a peer (and its
+ * service metadata) from cache — that just breaks every follow-up request
+ * pinned to the same peer/service. We only drop a peer once it has failed
+ * repeatedly within a short window.
+ */
+const PEER_FAILURE_THRESHOLD = 4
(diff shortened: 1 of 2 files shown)


kotevcode:
@claude wdyt?

Issue #355, opened 2026-04-24
Default ICE config is STUN-only — new users behind symmetric NAT / CGNAT / corp firewall cannot connect to any peer


kotevcode:
## Symptom

A fresh desktop-app user hit this immediately after install:

```
12:03:34 AM [connect] [system] AI stream interrupted. Retrying in 2.0s (1/3)
  [stream_error, non-retryable: The stream stopped before completion:
   502 P2P request failed: Connection to 1d90f467689d... failed]
```

The peer they were pinned to (`1d90f467…`, a deepinfra k8s pod) was serving
other buyers (`232292ef…`) successfully at the same time. The failure was
WebRTC ICE pair selection, not the seller being down.

## Root cause

`packages/node/src/p2p/ice-config.ts` returns **STUN-only** defaults:

```ts
export function getDefaultIceConfig(): IceConfig {
  return {
    iceServers: [
      { urls: "stun:stun.l.google.com:19302" },
      { urls: "stun:stun1.l.google.com:19302" },
      { urls: "stun:stun2.l.google.com:19302" },
    ],
    iceTransportPolicy: "all",
  };
}
```

We ship **no TURN relay anywhere in the default build**. `needsTurnFallback()`
is written but nothing populates the `turnServers` argument to
`buildIceConfig()`.

Consequences:
- Users on symmetric NAT, corporate wifi, hotel wifi, CGNAT, cellular, or many
  home routers cannot establish a DataChannel to any seller — ever.
- The k8s peers log `[NAT] UPnP/NAT-PMP mapping failed — seller may not be
  reachable from the internet` on startup, which is a direct symptom: they
  rely on GKE's NAT producing a usable srflx candidate, and when the buyer
  is *also* behind restrictive NAT the ICE pair never connects.
- First-time desktop downloads fail silently with opaque
  `Connection to <peerId> failed` errors.

## Fix

1. Stand up a TURN server (coturn on a small VM, or Twilio/Xirsys/Cloudflare
   Calls) and plumb credentials into `buildIceConfig` as the default.
2. Add a retry path that switches to `iceTransportPolicy: "relay"` if
   host/srflx pairs haven't connected within ~2s.
3. Make TURN config overridable via desktop settings / CLI `--ice-server` flag
   for self-hosted fleets.

## Related

- `packages/node/src/p2p/ice-config.ts`
- `packages/node/src/p2p/` (where the RTCPeerConnection is constructed)
[...]

PR #356, opened 2026-04-24
fix(buyer): Clearer errors when pinned peer is unreachable

kotevcode:
## Summary

A first-time Mac desktop user on a residential NAT hit a sequence of confusing 502s after install:

- `Connection to 1d90f467… failed` (real cause: WebRTC / NAT traversal, tracked in #355)
- `No peers support openai-chat-completions` (false — the pinned peer was simply missing from the DHT momentarily)
- `No peers support anthropic-messages` (false — the pinned peer could serve this via cross-protocol transformation)

The transport failure is being addressed separately by deploying TURN (#355). This PR fixes the cascade of misleading errors that show up after a transport failure, and makes the pinning path more lenient so cross-protocol transformation can still rescue a pinned peer whose service-name metadata misses the exact requested model.

### `apps/cli/src/proxy/buyer-proxy.ts` — explicit pinned-peer error paths

Collapsed two separate refresh-and-retry blocks into a single preflight that triggers on any of: pin missing from cache, candidate set empty, or cache past TTL. Split the old generic `No peers support X` branch into three ordered, actionable cases:

1. Pin not discoverable in DHT → `Pinned peer X is not reachable right now… Pick a different service in Discover or try again later.`
2. Pin discoverable but filtered by protocol/service → `Pinned peer X does not support this request (protocol=…, provider=…, service=…).`
3. Invariant fallback (should be unreachable given the checks above).

### `apps/cli/src/proxy/routing.ts` — `ServiceFilterMode`

New `'strict' | 'lenient'` parameter threaded through `getPeerProviderProtocols`, `resolvePeerRoutePlan`, and `selectCandidatePeersForRouting`. Strict stays the default (correct for future auto-routing). Buyer-proxy passes `'lenient'` since every request is pinned — when the pinned peer's advertised service names don't include the exact requested model, the peer still gets a route plan (via the cross-protocol adapter if needed) and the seller returns a truthful upstream model-not-found error instead of the buyer silently dropping the only peer it was asked to use.

### `apps/cli/src/proxy/buyer-proxy.test.ts` — lenient-mode test

New test exercises both modes against the same peer shape and verifies that lenient mode keeps the peer and selects the transformed protocol path. All 68 CLI tests pass.

## Related

- #355 — default ICE config is STUN-only; the transport issue that triggered this investigation.
- Commit b55c853 (`feat(cli): disable auto peer selection`) — removed the multi-peer retry loop, so errors from a single failed peer now surface directly to the user.

## Test plan

- `pnpm run typecheck` — root, clean
- `pnpm -C apps/cli test` — 68/68 pass
- Manual: a pinned peer offline now emits a peer-specific "not reachable" error instead of "No peers support X"


kotevcode:
@claude review this bro

PR #361, opened 2026-04-25
fix(node): Reliable peer lookup via per-peer DHT topic

kotevcode:
## Description

A single wildcard DHT topic (`antseed:*`) doesn't scale: as the network grew past the K closest storage nodes' capacity, individual peers became intermittently invisible — back-to-back lookups returned 15 / 61 / 60 endpoints with the same target dropping in and out. Stats pods, buyer daemons, and `antseed network peer <id>` all suffered.

This PR makes peer lookups deterministic and recoverable:

- **Per-peer DHT topic.** Each peer now also announces under `antseed:peer:<peerId>`, and `PeerLookup.findByPeerId(peerId)` / `AntseedNode.findPeer(peerId)` use it for a single-infohash lookup. Spoofing protection: only endpoints whose served metadata's `peerId` actually matches the request are returned, and the result cap is applied after that identity filter.
- **Tighter reannounce cadence.** Default 15 min → 5 min (well under the ~30 min BEP-5 storage TTL that caused the visibility gaps).
- **Longer DHT operation timeout.** 10 s → 25 s so cold routing tables finish fanning out before lookups resolve.
- **Higher lookup result cap.** `maxResults` 50 → 200 so `network browse` stops silently truncating on a busy network.
- **CLI** `network peer <id>`: cache miss now tries the per-peer topic first (deterministic), then falls back to a wildcard scan only for peers still on older builds. This also fixes the cache-present/cache-miss case where the old command skipped live DHT discovery.
- **`scripts/find-peer.mjs`**: standalone fresh-DHT diagnostic script that bypasses local caches, tries the per-peer topic first, and uses wildcard only as a fallback/old-path comparison.

## Review Follow-up

- Hardened `PeerLookup.findByPeerId()` to reject malformed peer IDs before touching the DHT.
- Moved exact-peer metadata filtering ahead of the `maxResults` cap so topic-squatting endpoints cannot fill the cap before the honest endpoint is considered.
- Updated the diagnostic script to exercise the new per-peer lookup path first.

## Release Notes

- `antseed network peer <peerId>` now uses a deterministic per-peer DHT topic for lookup, so peers are reliably found by id even when the wildcard topic is saturated.
- Sellers reannounce every 5 minutes (down from 15) so brief DHT hiccups no longer make a peer disappear from buyers' views.
- `antseed network browse` no longer silently truncates results at 50 peers.

## Types of Changes

## Checklist

### DIFF
--- a/apps/cli/src/cli/commands/network/bootstrap.ts
+++ b/apps/cli/src/cli/commands/network/bootstrap.ts
@@ -2,7 +2,7 @@ import type { Command } from 'commander'
 import chalk from 'chalk'
 import { getGlobalOptions } from '../types.js'
 import { loadOrCreateIdentity } from '@antseed/node'
-import { DHTNode, parseBootstrapList } from '@antseed/node/discovery'
+import { DEFAULT_DHT_CONFIG, DHTNode, parseBootstrapList } from '@antseed/node/discovery'
 import { setupShutdownHandler } from '../../shutdown.js'
 
 export function registerNetworkBootstrapCommand(networkCmd: Command): void {
@@ -40,8 +40,8 @@ export function registerNetworkBootstrapCommand(networkCmd: Command): void {
         peerId: identity.peerId,
         port,
         bootstrapNodes,
-        reannounceIntervalMs: 15 * 60 * 1000,
-        operationTimeoutMs: 10_000,
+        reannounceIntervalMs: DEFAULT_DHT_CONFIG.reannounceIntervalMs,
+        operationTimeoutMs: DEFAULT_DHT_CONFIG.operationTimeoutMs,
       })
 
       await dht.start()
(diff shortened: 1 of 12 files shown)


kotevcode:
@claude review this

kotevcode:
Addressed the Claude review follow-ups: hardened direct PeerLookup input validation, moved exact-peer metadata filtering ahead of the result cap, and updated scripts/find-peer.mjs to exercise the per-peer topic first with wildcard only as fallback diagnostics. Also ran focused node tests/typechecks and an independent subagent review.

kotevcode:
Backward compatibility follow-up: new sellers remain visible to old buyers because wildcard/service announcements are unchanged. New buyers now also remain compatible with old sellers: AntseedNode.findPeer() tries the per-peer topic first, then falls back to wildcard discovery when the seller has not upgraded yet. Added find-peer compatibility tests and pushed 9a3258e.

PR #385, opened 2026-04-26
fix(node): shard peer discovery across subnet topics

kotevcode:
## Description

The single wildcard topic concentrates every announcer onto one DHT infohash; once the K-closest storage nodes saturate, lookups silently return inconsistent subsets and peers stay invisible until the buyer restarts. This shards announcements across `SUBNET_COUNT = 16` deterministic subnet topics (`antseed:subnet:<i>`, where `i = firstByte(peerId) % 16`) and fans `PeerLookup.findAll()` out across them in parallel. The wildcard topic is kept on both sides as a transition fallback so older sellers still on the single-topic build remain visible to upgraded buyers.

## Why restart helped before

A long-running buyer's DHT routing table stabilises onto one K-closest neighbourhood for the wildcard infohash and keeps missing the same peers. A fresh bootstrap finds a different K-closest set, which is why closing and reopening the buyer surfaced peers that had been invisible.

## Changes

- `packages/node/src/discovery/dht-node.ts` — `SUBNET_COUNT`, `subnetOf(peerId)`, `subnetTopic(i)` with rationale documenting the K-closest saturation problem and why 16 is the right shard count for the foreseeable network size.
- `packages/node/src/discovery/announcer.ts` — sellers announce on their deterministic subnet topic in addition to the wildcard and per-peer topic.
- `packages/node/src/discovery/peer-lookup.ts` — `findAll()` fans out across all 16 subnets plus the wildcard via `Promise.allSettled`, so a single misbehaving subnet lookup can no longer black-hole the whole enumeration. Existing `host:port` dedup absorbs the overlap from sellers announcing on both subnet and wildcard.
- `apps/network-stats/src/poller.ts` — same fan-out applied so the public network-stats observer doesn't suffer the same blind-spot.
- `scripts/find-peer.mjs` — diagnostic now tries per-peer topic → peer's expected subnet → wildcard, in that order.

## Tests

`@antseed/node` — 588 / 588 pass (6 net new):

- `dht-topic.test.ts` — subnet topic format, casing / `0x` stability, **exact-uniformity** distribution check (asserts `256 % SUBNET_COUNT === 0` so future skew-inducing values trip), malformed-input fallback.
- `peer-lookup.test.ts` — fan-out shape, transition-phase dedup (subnet + wildcard collapsing to one metadata resolve), **rejection resilience** (one subnet lookup throws → others still surface), **announcer ↔ lookup symmetry** (exact hash-set equality), **property-style coverage** (every possible `subnetOf(peerId)` is in the lookup's queried set).
- `announcer-load.test.ts` — count assertion now includes the subnet announce alongside service / wildcard / per-peer.

`pnpm run typecheck` clean across all 19 packages. The pre-existing `@antseed/desktop` test failure on `main` is unrelated.

## Compatibility

- **Phase 1 (this PR):** sellers announce both subnet and wildcard; buyers query both. New buyers see everyone, old buyers still see new peers.
- **Phase 2 (later, after sellers upgrade en masse):** drop wildcard announce + lookup as a one-line follow-up.


kotevcode:
@claude review this pr

kotevcode:
@claude i removed service topics completely, should solve 2 but also was not in use. review again.

PR #407, opened 2026-04-27
Fix discover chat peer binding

kotevcode:
## Summary

- Preserve the selected Discover peer/provider when creating a new chat conversation.
- Persist the peer binding into the session at creation time so conversation summaries and the chats list immediately show the correct peer.
- Include peer identity in the service-options signature so peer-only catalog changes cannot leave stale selections.

## Tests
- pnpm --filter=@antseed/desktop run build:main
- pnpm --filter=@antseed/desktop run typecheck:renderer
- pnpm --filter=@antseed/desktop run build:renderer
- node --test apps/desktop/dist/main/*.test.js

Note: `pnpm --filter=@antseed/desktop run test` still fails because its script invokes `node --test dist/main`, and Node cannot resolve the directory entrypoint. Running the compiled test files directly passes

### DIFF
--- a/apps/desktop/src/main/pi-chat-engine.ts
+++ b/apps/desktop/src/main/pi-chat-engine.ts
@@ -1514,6 +1514,8 @@ class PiConversationStore {
         usage: conversation.usage,
         totalTokens,
         totalEstimatedCostUsd: deriveCost(conversation.messages),
+        ...(conversation.peerId ? { peerId: conversation.peerId } : {}),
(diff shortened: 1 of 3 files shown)


kotevcode:
Follow-up: fixed Discover card generated descriptions so Mistral-family models (including Magistral/Ministral) are no longer described as only Mistral's flagship model, and model copy consistently includes the serving peer.\n\nAdditional checks:\n- pnpm --filter=@antseed/desktop run typecheck:renderer\n- pnpm --filter=@antseed/desktop run build:renderer

PR #415, opened 2026-04-28
fix(desktop): stop leaking antseed-proxy as upstream provider

kotevcode:
## Summary

Fixes a regression where the desktop app's pi-chat-engine would persist the local proxy sentinel (`antseed-proxy`) as a conversation's upstream provider whenever no real provider was known, then re-emit it on every subsequent send as the `x-antseed-provider` HTTP header. The buyer proxy matches that header against the pinned peer's advertised providers and rejects unknown values, producing the confusing user-visible error:

> 502 Pinned peer 0e49122e76bd... does not offer provider=antseed-proxy. Available providers: openai. Remove or change the x-antseed-provider header, or pick a different peer.

`PROXY_PROVIDER_ID` ("antseed-proxy") is an internal label the desktop's pi-ai SDK config uses to wire up the OpenAI/Anthropic SDK to the local buyer proxy. It must never leave the machine.

## Changes

- New module `apps/desktop/src/main/chat-provider-hint.ts` exporting `PROXY_PROVIDER_ID`, `normalizeProviderId`, and `sanitizeProviderHint` (drops empty values **and** the local proxy sentinel).
- `pi-chat-engine.ts` applies the rule at all three boundaries:
  - **Write path** (`PiConversationStore.create`): persist `''` instead of the sentinel when no real provider is known.
  - **Read path** (`buildConversationFromManager`): existing sessions with the sentinel come back as no-hint, so old conversations heal automatically.
  - **Wire boundary** (`makeProxyService`): defence-in-depth re-sanitize before setting the `x-antseed-provider` header.

Without a hint, the buyer proxy auto-selects a peer that supports the service.

## Test plan

- [x] New unit tests in `chat-provider-hint.test.ts` cover the sentinel-stripping and the preserved real-provider cases (`openai`, `anthropic`, `local-llm`, plus case/whitespace/null/undefined edges).
- [x] `npm --prefix apps/desktop run build:main` — clean.
- [x] `npm --prefix apps/desktop run typecheck:renderer` — clean.
- [x] `node --test apps/desktop/dist/main/*.test.js` — all 66 desktop main tests pass (5 new + 61 existing).

https://claude.ai/code/session_01SFHJYVP65g5MaA1vL4JGXs

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFHJYVP65g5MaA1vL4JGXs)_

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@antseed/desktop",
   "productName": "AntSeed Desktop",
-  "version": "0.1.63",
+  "version": "0.1.64",
   "private": true,
   "description": "Desktop app for AntSeed",
   "author": "AntSeed <hello@antseed.com>",
--- a/apps/desktop/src/main/chat-provider-hint.test.ts
+++ b/apps/desktop/src/main/chat-provider-hint.test.ts
@@ -1,0 +1,44 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+import {
+  PROXY_PROVIDER_ID,
+  normalizeProviderId,
+  sanitizeProviderHint,
+} from './chat-provider-hint.js';
+
+test('normalizeProviderId trims and lowercases real provider names', () => {
+  assert.equal(normalizeProviderId(' OpenAI '), 'openai');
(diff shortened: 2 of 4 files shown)


kotevcode:
@claude

PR #416, opened 2026-04-28
fix(desktop): persist real upstream provider on every chat send

kotevcode:
Closing in favour of a much simpler fix — the `x-antseed-provider` header isn't actually required when peer is pinned (the buyer proxy resolves the route plan from peer + service ID alone). Replacement PR will just stop emitting the header.


kotevcode:
@claude

PR #417, opened 2026-04-28
fix(desktop): stop sending x-antseed-provider header

kotevcode:
## Summary

Replaces #416 with the minimal fix: **stop sending `x-antseed-provider`**. The buyer proxy resolves the route plan for a pinned-peer request from the peer + the service ID in the request body alone (`apps/cli/src/proxy/buyer-proxy.ts:1093-1108` only checks `explicitProvider` when it's non-null). Antstation always pins the peer, so the header was redundant — and the only thing that gave internal SDK labels a way to leak onto the wire.

## Changes

- `makeProxyService` no longer sets `x-antseed-provider`. Only `x-antseed-pin-peer` and `x-antseed-spending-auth` go out.
- `resolveProviderHintForService` removed (now unused).
- `runStreamingPrompt` drops its `providerOverride` parameter.
- IPC handlers `chat:ai-send-stream` / `chat:ai-send` still accept (and ignore) the renderer's `provider` arg, for ABI compatibility with older renderers.
- Bumps `apps/desktop` to 0.1.65.

#415's read-path / write-path sanitize stays in place as belt-and-suspenders.

Net: 2 files changed, 15 insertions, 21 deletions.

## Test plan

- [x] `npm --prefix apps/desktop run build:main` — clean.
- [x] `npm --prefix apps/desktop run typecheck:renderer` — clean.
- [x] `node --test apps/desktop/dist/main/*.test.js` — 66/66 passing.

https://claude.ai/code/session_01SFHJYVP65g5MaA1vL4JGXs

---
_Generated by [Claude Code](https://claude.ai/code/session_01SFHJYVP65g5MaA1vL4JGXs)_

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@antseed/desktop",
   "productName": "AntSeed Desktop",
-  "version": "0.1.64",
+  "version": "0.1.65",
   "private": true,
   "description": "Desktop app for AntSeed",
   "author": "AntSeed <hello@antseed.com>",
--- a/apps/desktop/src/main/pi-chat-engine.ts
+++ b/apps/desktop/src/main/pi-chat-engine.ts
@@ -420,12 +420,6 @@ function updateServiceProtocolMap(
   }
 }
 
-function resolveProviderHintForService(
-  explicitProvider?: string,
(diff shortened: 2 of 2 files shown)


kotevcode:
@claude

PR #423, opened 2026-04-29
Update DIEM staking hero and live capacity

Amosme:


Augustas11:
The hero refresh looks great and the capacity-replay logic in `useVeniceCapacityUsage24h` (`lib/hooks.ts:142–230`) is exactly the right model: replay the same-day timeline of `Staked` and `ChannelSettled` events ordered by `(blockNumber, logIndex)`, with `initial left = current capacity − same-day adds`. That correctly handles a user staking mid-day after an earlier settlement.

A few things worth addressing:

1. **Hardcoded `$0.50` DIEM price.** `lib/hooks.ts:208, 215` use `Number(amount) / 1e18 * 0.5`. The app already exposes `useDiemPrice()` via Coinbase or the Diem oracle, so a live price would make the "Capacity / Left" tile track market reality. If $0.50 is intentionally the protocol's reference price (vs market), please add a code comment citing the source so the next reader doesn't change it.

2. **Binary search for the reset block runs ~27 `getBlock` calls per cold load.** The `useQuery` `staleTime: 60_000` caches it for a minute, but every browser refresh / route mount re-runs the search. Caching the resolved `fromBlock` keyed by `(resetStart, headBlock // ~1000)` in `localStorage` would shave the cost dramatically. Even just memoizing across mounts in a module-level Map would help.

3. **Large CSS removal — please run a visual regression sweep.** The diff strips ~400 lines of CSS (`.panel-intro`, `.claim-stat`, `.unstake-flow-card`, etc.) and compacts many other rules. Some of these classes are still referenced by JSX in the same file. `git diff --check` doesn't catch this. A manual walkthrough of all four tabs (Stake, Unstake, Claim, plus the unstake state machine views) before merging would be cheap insurance.

4. **`Staked` event** added to `abi.ts:36`. Worth verifying the deployed `DiemStakingProxy` (`0x1f228613116E2d08014DfdCC198377C8dedf18C9` per HOTSPOTS H17) actually emits this event. If the `getLogs(Staked)` call returns nothing, capacity-left math silently uses initial-only values.

5. **`apps/diem-staking` has zero tests** (HOTSPOTS H17 — handles real money). This PR adds a non-trivial timeline-replay function — a perfect candidate to seed the test pattern. A 30-line unit test feeding synthetic stake + settle logs to `useVeniceCapacityUsage24h`'s inner loop would catch ordering/casting regressions.

6. **Lost rationale comments.** `Layout.tsx:9-12` removed the comment explaining why the OS glyph matches `apps/website/src/lib/DesktopDownloadIcon.tsx`. That kind of cross-reference comment is exactly what saves the next reader. Consider keeping them.

No invariants or protocol surface touched — this is a display-layer PR.

Amosme:
Did you add it ?  Can’t see it

PR #439, opened 2026-05-02
Fix assistant response boundaries and streaming anchor

### LINKED ISSUE #424: AntStation chat: assistant responses have no visual boundary, hard to find where answer starts
## Problem

In AntStation's chat, you can't tell where the assistant response starts. After hitting send, the answer streams in but there's no visual anchor for "this is the start of the reply" — the text just flows into the layout. Users end up scrolling up to find their own prompt as a reference point for where the answer began, which is a poor read experience.

WhatsApp and Anthropic's Cowork both solve this by giving every message — user and assistant — a clear visual container.

## Root cause: asymmetric message styling

User messages and assistant messages are styled completely differently in [`apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss`](apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss):

```scss
&.own {                                    // user message
  width: fit-content;
  margin-left: auto;                       // right-aligned
  background: var(--chat-user-bubble);     // colored
  border: 1px solid …;                     // bordered
  border-radius: var(--radius);            // rounded
  padding: 9px 17px;
  margin-bottom: 32px;
}

&.other {                                  // assistant message
  align-self: stretch;
  background: transparent;                 // no fill
  border-left: 0;
  border-radius: 0;
  width: 100%;
  padding: 0 16px;
}
```

User messages get a bubble. Assistant messages get full-width, transparent, no border, no avatar, no role label — they read as raw page content rather than a reply. With the auto-scroll behavior in [`ChatView.tsx:176-214`](apps/desktop/src/renderer/ui/components/views/ChatView.tsx#L176-L214) keeping the view pinned to the bottom of the streaming response, the user's own prompt and the start of the assistant reply are off-screen — and there's no visual marker to scroll back to.

## Suggested fixes

Pick one or combine:

1. **Give assistant messages a container too.** A subtle background tint, a left accent bar, or a card shadow — something that bounds the message visually. Doesn't need to be heavy; just enough that "where the reply starts" is obvious at a glance.
2. **Add a small role header.** Avatar + name (e.g. ant icon + model display name) at the top of each assistant message. WhatsApp does avatars; Claude/Cowork does an avatar + label. This becomes the visual anchor users scan for.
3. **Auto-scroll to the start of the response, not the end.** When a new assistant message begins streaming, scroll so the message header is at the top of the viewport, not so the latest token is at the bottom. This is what ChatGPT does and the reason users don't get lost there.
4. **Visible separator between turns.** A faint divider line + extra vertical spacing between the end of one assistant message and the start of the next user prompt. Cheap, helps a lot.

#3 is probably the biggest single quality-of-life win — it directly addresses "I have to scroll up to find my answer."

## User impact

Reading a long response feels like landing in the middle of a document. New users especially can't tell at a glance which text is the model's answer vs. UI chrome vs. the previous turn. Easy to fix and a meaningful polish item.

SheferKagan:
## Summary

This PR fixes the chat readability problem from #424 by making assistant turns visibly bounded and by anchoring new streaming replies at their start instead of immediately chasing the newest token.

### What changed

- Added a subtle assistant/system response card:
  - light card fill
  - 1px border
  - rounded corners
  - soft shadow
  - green left accent edge
  - extra turn spacing
- Added a persistent response header for non-user messages:
  - compact avatar/initial
  - `Assistant` / `System` label
  - existing metadata stays visible on completed responses
  - header appears while streaming too, so users immediately see where the reply begins
- Added a dedicated chat bubble body wrapper so long markdown/tool output remains contained inside the card.
- Improved streaming scroll behavior:
  - when a new assistant response starts, the chat scrolls to the top of that response card
  - the app no longer force-scrolls every streaming token to the bottom from the chat module
  - manual scrollback is still respected
  - if the user scrolls back to the bottom, normal live bottom-follow behavior resumes
- Kept user bubble styling, copy actions, markdown rendering, tool groups, and existing completed-response metadata behavior intact.

## Why

AMOS reported that assistant responses currently read like raw page content: no border, no fill, no header, and no visual anchor. With bottom-pinned streaming, long responses can leave the user looking at the middle/end of an answer with no clear indication where it started.

This PR addresses both sides:

1. visual boundaries make every assistant answer scannable as a distinct turn;
2. response-start anchoring keeps the start of a new answer visible when streaming begins.

Fixes #424

## Files changed

- `apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss`
  - styles assistant cards, header, avatar, body wrapper, and tool header background inside the new card
- `apps/desktop/src/renderer/ui/components/chat/ChatBubble.tsx`
  - renders the assistant/system header and exposes a streaming anchor marker
- `apps/desktop/src/renderer/ui/components/views/ChatView.tsx`
  - anchors new streaming responses at the start of the response card while preserving manual scroll behavior
- `apps/desktop/src/renderer/modules/chat.ts`
  - removes the old module-level streaming token bottom-scroll queue so the renderer can own the new scroll policy

## Checks

- `pnpm --filter=@antseed/desktop run typecheck:renderer` ✅
- `pnpm --filter=@antseed/desktop run build:renderer` ✅
- `pnpm --filter=@antseed/api-adapter run build` ✅
- `pnpm --filter=@antseed/node run build` ✅
- `pnpm --filter=@antseed/payments run build:server` ✅
- `pnpm --filter=@antseed/desktop run build:main` ✅
- `pnpm --filter=@antseed/desktop test` ✅ 66 passing tests

Note: `desktop test` emits expected npm `.npmrc` warnings and expected test log lines for mocked network-stats failure cases; all tests pass.

## Merge note

Do not merge until explicitly approved by maintainers.

### DIFF
--- a/apps/desktop/src/renderer/modules/chat.ts
+++ b/apps/desktop/src/renderer/modules/chat.ts
@@ -811,27 +811,6 @@ export function initChatModule({
     updateStreamingIndicator();
   }
 
-  // ---------------------------------------------------------------------------
-  // Scroll helper
-  // ---------------------------------------------------------------------------
-
-  function scrollChatToBottom(): void {
-    const container = document.querySelector<HTMLElement>('[data-chat-scroll]');
-    if (!container) return;
-    const threshold = 100;
-    const distanceFromBottom =
-      container.scrollHeight - container.scrollTop - container.clientHeight;
-    if (distanceFromBottom < threshold) {
-      container.scrollTop = container.scrollHeight;
-    }
-  }
-
-  function queueScrollChatToBottom(): void {
-    requestAnimationFrame(() => {
-      scrollChatToBottom();
-    });
-  }
-
   function cloneStreamingMessage(message: ChatMessage): ChatMessage {
     return {
       ...message,
@@ -858,7 +837,6 @@ export function initChatModule({
   function setStreamingMessage(message: ChatMessage | null): void {
     uiState.chatStreamingMessage = message ? cloneStreamingMessage(message) : null;
     notifyUiStateChangedSync();
-    if (message) queueScrollChatToBottom();
   }
 
   function getConversationStreamingMessage(convId: string): ChatMessage | null {
--- a/apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss
+++ b/apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss
@@ -258,12 +258,20 @@
 
   &.other {
     align-self: stretch;
-    background: transparent;
+    background: color-mix(in srgb, var(--bg-card) 86%, var(--chat-area-bg));
     color: var(--text-primary);
-    border-left: 0;
-    border-radius: 0;
-    width: 100%;
-    padding: 0 16px;
+    border: 1px solid var(--border);
+    border-left: 3px solid rgba(var(--accent-rgb), 0.55);
+    border-radius: var(--radius);
+    box-shadow: var(--shadow-sm);
+    width: calc(100% - 32px);
+    padding: 12px 16px 10px;
+    margin: 4px auto 28px;
+
+    &:focus-within {
+      border-color: rgba(var(--accent-rgb), 0.32);
+      border-left-color: rgba(var(--accent-rgb), 0.7);
+    }
 
   :global {
     /* Tool group */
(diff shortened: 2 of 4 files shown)


SheferKagan:
Closing as no longer relevant. The chat direction is moving toward the assistant turn activity separation foundation and stacked UI exploration instead.

PR #452, opened 2026-05-04
feat(network-stats): web dashboard SPA with peer detail, charts, and chain backfill

owklama:
## Summary

Builds out the `apps/network-stats/` HTTP service end-to-end: a Vite + React
SPA on top of a modular Express backend, three new SQLite tables, a one-shot
chain backfill, and the derived-analytics layer (`/insights`) that powers the
dashboard's leaderboards, pricing market, and velocity panels.

### Frontend (`apps/network-stats/web`)

- Vite + React + TanStack Query SPA served from `dist/web` in production; in
  dev, Vite runs on 5180 and proxies `/stats` and `/health` to the backend.
- Sidebar nav splits Network / Peers / Insights / Leaderboards / Price
  movements; per-peer detail view reachable from the swarm table.
- Two recharts time-series (active peers + requests, tokens served) backed by
  `/history`, plus a backfill-progress strip.
- Polls the narrower split endpoints (see below) on the routes that need
  them; falls back to the unions only where convenient.

### Backend HTTP layer

Server is a composition root only (94 lines): it wires DI, middleware, and
the start/stop lifecycle. Per-endpoint logic lives under `src/http/routes`
(stats, history, insights, health) with shared infrastructure
(`agentIdCache`, `ttlCache`, validators, serializers, CORS + error
middleware) alongside.

**Endpoints** — each union is kept for back-compat; the split endpoints exist
so the frontend can poll cheap data at a higher cadence than the heavy
fan-outs.

- `GET /stats` — union (legacy)
- `GET /stats/network` — DHT-only metrics + indexer health + backfill state (cheap)
- `GET /stats/peers` — per-peer DHT snapshot enriched with on-chain totals (heavy)
- `GET /insights` — union (legacy)
- `GET /insights/leaderboards` — most active / settlements / buyers / staked /
  diverse / newest / oldest, plus `trendingUp` / `trendingDown` (24h vs prior-7d
  daily-average ratio, 5-req floor)
- `GET /insights/pricing` — `byService` distribution (min/median/p90/max + cheapest peer),
  `priceStability` (`mostStable` / `mostVolatile`), `priceMovers` (`biggestDrops`
  / `biggestHikes`) over a 30d window
- `GET /insights/services` — top services / categories / protocols / providers,
  per-region peer counts, and concentration (Gini / Herfindahl / top-10 share)
- `GET /insights/activity` — live peers online + sellers active in 24h,
  plus `velocity.last24h` / `velocity.last7d` deltas + growth %
- `GET /history?range=1d|7d|30d` — union (legacy); 1h buckets for `1d`, 1d
  buckets for `7d`/`30d`. Gauges use last-sample-in-bucket; counters are
  per-bucket deltas. 400 on bad range; empty payload when the store isn't
  configured.
- `GET /history/peers` — same shape, peer-activity fields only
- `GET /history/tokens` — same shape, tokens fields only
- `GET /health` — liveness probe

`/insights` derivations are 15s-memoized in-process (server-instance keyed)
since the underlying indexer state advances at minute cadence at best.

### Storage

Three new SQLite tables in the network-stats store:

- **`network_history`** — minute-cadence time-series feeding `/history` and
  `/insights.velocity`. Sampler is gated on the chain backfill resolving so
  cumulative counters stay monotonic in `ts` order.
- **`peer_pricing_history`** — per-(peer, provider, service) pricing samples
  written on every successful poll. Dedups on equal prices via INSERT OR
  IGNORE; a daily heartbeat keeps stable peers visible in the volatility
  window. Backs `priceStability` + `priceMovers`.
- **`seller_activity_history`** — per-minute per-seller cumulative-totals
  snapshots. Two filters before the SQL fires: skip sellers dormant past the
  trending window, and skip sellers whose `(totalRequests | totalInputTokens
  | totalOutputTokens | settlementCount)` signature matches the last sample.
  Backs `leaderboards.trendingUp` / `trendingDown`.

### Backfill + sampler ordering

- One-shot chain backfill replays `MetadataRecorded` events from the
  contract's deploy block (or the existing checkpoint, if resuming) through
  the same `store.applyBatch` path the live indexer uses, then writes one
  `network_history` row per UTC day from `getNetworkTotals()`. Skipped when
  ≥6h of forward history already exists. Reuses the `StatsClient`'s provider
  so any FallbackProvider failover also covers backfill; chunk size +
  inter-chunk delay are env-tunable to stay within public-RPC rate limits.
- Backfill and the live indexer share the same `applyBatch` path, so they
  must not run concurrently — the indexer is now started in the backfill's
  `.finally` (success, failure, or skipped), not in parallel.
- Live `ts=now` sampling logic is encapsulated in a `HistorySampler` class
  (extracted out of `index.ts`), which collapses three coordination concerns
  into one place:
  1. DHT-not-yet-observed → write the `ACTIVE_PEERS_UNKNOWN` sentinel
     instead of a misleading zero
  2. Backfill-vs-now ordering → defer all `ts=now` writes until backfill
     resolves so cumulative counters stay monotonic
  3. Per-seller dedup → skip the SELECT+INSERT against
     `seller_activity_history` when a seller's totals haven't changed

### Supporting changes

- **`NetworkPoller.onPollComplete(snapshot)`** — opaque callback so wiring
  stays storage-agnostic; `index.ts` is the one place that knows about both
  the poller and the store.
- **`metrics` module** (renamed from `aggregates`) — pure, DHT-only network
  metrics computed on every `/stats` hit.
- **`insights` module** — pure derivations on the `/insights` payload.
  Server does the I/O, hands arrays/maps in; module is fully fixture-tested.
- **`utils.ts`** — shared helpers (`mapWithConcurrency`, `normalizeAddress`,
  `getPeerLookupAddress`, `bigintDesc`, `gini`, `herfindahl`, `topShare`,
  `distribution`, block-timestamp resolution, in-place progress reporter).
- **`.env.local` auto-load** — package-root `.env.local` is loaded on
  startup via `process.loadEnvFile`, keeping an Alchemy/QuickNode
  `NETWORK_STATS_RPC_URL` out of the shell. Silent when the file is missing.
- **Root `pnpm dev:network-stats` script** — builds `@antseed/node` then
  runs `vite`, `tsc --watch`, and `node --watch` concurrently.

## Test plan

- [ ] `pnpm --filter=@antseed/network-stats run typecheck` passes
- [ ] `pnpm --filter=@antseed/network-stats run test` passes — covers store
      bucketing + price/activity rollups, `/history` + `/insights` shapes
      (union + split), `/insights` cache TTL, backfill chunking + progress
      reporting, the `applyBatch`-shared backfill path, the metrics +
      insights pure modules, and the new route modules under `http/`
- [ ] `pnpm --filter=@antseed/network-stats run build` produces `dist/`
- [ ] `pnpm dev:network-stats` — `/stats/network` exposes `backfill` +
      `indexer` health, `/stats/peers` returns enriched peer rows,
      `/history/peers` and `/history/tokens` return bucketed points,
      `/insights/leaderboards` + `/insights/pricing` + `/insights/activity`
      populate the dashboard panels, sampler stays paused until backfill
      resolves
- [ ] Re-running with ≥6h of history skips the chain backfill (logged)


Augustas11:
This is impressive work — the sampler architecture, the backfill-then-live ordering, and the per-route endpoint splits are all the right design choices. A few comments before this lands:

1. **PR size is a real review hazard.** 65 files / +15,706 / -528 in one merge is too large to safely review line-by-line. The five thematic chunks called out in the body would each be a complete reviewable PR on its own:
   - (a) `HistorySampler` + per-seller dedup (sampler.ts + indexers/)
   - (b) one-shot backfill machinery + tests (`indexers/backfill.ts`, `backfill.test.ts`)
   - (c) `/insights` derivations + 15s response cache (`insights.ts`, `http/response-cache.ts`)
   - (d) endpoint splits + composition-root server (`http/routes/`, `server.ts`)
   - (e) Vite + React SPA (`web/`)
   Even a 2-way split (server + frontend) would help future bisecting and rollback.

2. **Coordination with PR #318.** That PR adds a `settlement_events` table and a `SettlementIndexer` for a `/leaderboard` endpoint. This PR adds a `seller_activity_history` table and a `HistorySampler` for `/insights/leaderboards.{trendingUp,trendingDown}`. The two are solving overlapping problems with different schemas. Whichever lands second will need a real merge: either pick one (probably the broader sampler architecture here) and rework #318's leaderboard on top of it, or carry both schemas with intent. Worth a sync with @kotevcode.

3. **Test plan: 5 of 6 items unchecked.** Specifically, the dev-server smoke tests and the "≥6h history skips backfill" assertion. Those are exactly the cross-cutting behaviors that benefit from manual verification on top of unit tests. Please run them and report before merge.

4. **Sampler `.finally` ordering vs SIGTERM.** When the process gets SIGTERM mid-backfill, does the indexer never start, or does it start in the cleanup handler? A small test that aborts a stub backfill and asserts no `setInterval` leaks would lock this in.

5. **`/insights` 15s cache during indexer crash.** Users will see stale data with no signal. Worth surfacing `cacheAgeMs` or `lastIndexerTickAt` somewhere in `/stats/network` so the dashboard can warn ("data is N minutes old, indexer hasn't ticked").

6. **Sampler dedup signature** — `(totalRequests | totalInputTokens | totalOutputTokens | settlementCount)` — what if a counter rolls back due to a deep reorg? The signature would briefly equal an older sample's, suppressing the new write. Probably fine since reorgs are bounded by `reorgSafetyBlocks`, but worth a test.

7. **`peer_pricing_history` daily heartbeat** — please document whether the heartbeat fires at UTC midnight, local time, or a 24h sliding window. Affects how operators interpret the `mostStable` / `mostVolatile` rankings.

No invariant or protocol surface touched (pure indexer + analytics).

PR #460, opened 2026-05-06
Add timeframe rollups and seller rankings to network-stats

owklama:
Adds time-windowed activity rollups (24h / 7d / 30d) and four global seller ranking lists to the `/network-stats` API.

## Schema
New `seller_settlement_events` table — one row per `MetadataRecorded` event, written only when the indexer has the block timestamp. Indexed on `(agent_id, block_timestamp)` and `block_timestamp`. All-time totals continue to live in `seller_metadata_totals`.

## New data points

**`networkStats.windows.{last24h, last7d, last30d}`** — per-window aggregates over `seller_settlement_events` where `block_timestamp >= now − Nd`:
- `totalRequests = Σ request_count`
- `totalInputTokens = Σ input_tokens`
- `totalOutputTokens = Σ output_tokens`
- `settlementCount = COUNT(events)`

**`rankings.mostUsed[window]`** — sellers sorted by `totalRequests` desc.

**`rankings.topVolume[window]`** — sellers sorted by `totalInputTokens + totalOutputTokens` desc.

**`rankings.mostReach`** — all-time, sorted by `uniqueBuyers` desc (from `COUNT(DISTINCT buyer)` per agent in `seller_buyer_totals`). Excludes agents with 0 buyers.

**`rankings.risingStars`** — sellers whose recent rate exceeds their lifetime rate:
```
daysActive    = max(1, (now − firstSeenAt) / 86400)
lifetimeRate  = totalRequests / daysActive
score         = requests7d / lifetimeRate
```
Excluded: `totalRequests < 5`, `firstSeenAt = null`, `requests7d = 0`. Sorted by `score` desc — score > 1 means above lifetime average.

## Notes
- Window totals computed in JS with BigInt to preserve precision (SQL `SUM` on TEXT-encoded bigints would overflow or lose precision via REAL).
- Three SQL queries per request regardless of peer count — windows are computed once and reused across rankings.
- Default `NETWORK_STATS_MAX_BLOCKS_PER_TICK` raised from 2000 → 10000 (Alchemy comfortably handles this for the AntseedStats contract).


Augustas11:
The `BlockRangeIndexer` strategy abstraction is the right move — both `MetadataIndexer` and `ChannelsIndexer` now share the tick orchestration, reentrancy guard, and timestamp fetch in one place. And the BigInt-in-JS approach for window totals (per the PR body) is exactly the precision fix #318's leaderboard query needs (`store.ts:756–820` there does `SUM(CAST AS INTEGER)` and loses precision at the SQL→JS boundary).

Three things to address:

1. **Schema collision with PR #318.** Both PRs add per-event tables tied to `AntseedChannels` events. #318's `settlement_events` (from `ChannelSettled`) and this PR's `seller_settlement_events` (from `MetadataRecorded`) overlap heavily. #318 carries `platform_fee`, `buyer`, `tx_hash` for tx-correlation; this PR has the cleaner sampler abstraction but a narrower schema. Whoever lands second will need a real merge: I'd suggest carrying this PR's `BlockRangeIndexer` + #318's wider schema. Worth a sync with @kotevcode.

2. **Provider duplication in `BlockRangeIndexer.constructor` (`block-range-indexer.ts:153`).** It builds a fresh `ethers.JsonRpcProvider(options.rpcUrl)` — but `IndexerStrategy.fetchEvents` already calls through whatever provider the strategy holds, which may be a `FallbackProvider` from `BaseEvmClient`. Two providers = (a) double the RPC connections, (b) divergent failover behavior (a primary-RPC failure during `fetchEvents` falls over via Fallback, but a primary-RPC failure during `getBlock` here just errors). Either accept an `AbstractProvider` in the options, or pull it from the strategy via a `getProvider()` method.

3. **`NETWORK_STATS_MAX_BLOCKS_PER_TICK` default raise from 2000 → 10000.** Matches Alchemy comfortably, but public Base RPCs (and several free tiers) reject `eth_getLogs` ranges that large. Either keep the default at 2000 and document that operators on Alchemy-class RPCs can bump it, or land the bump alongside an explicit `RPC_TIER` config that picks the right default.

Smaller things:

4. **`risingStars` exclusions are correct** (`totalRequests < 5`, `firstSeenAt = null`, `requests7d = 0`) and the `daysActive < 7` natural damping prevents brand-new sellers from artificially topping the chart. Worth pinning with a unit test (very-new seller with 5 requests should rank near 1.0, not infinity).

5. **`getChannelEvents` decodes both `ChannelSettled` and `ChannelClosed`** into a `DecodedChannelEvent` union. Verify the `ChannelClosed` ABI matches `AntseedChannels.sol:311` exactly: `event ChannelClosed(bytes32 indexed channelId, address indexed buyer, address indexed seller, uint128 settled, uint128 refunded)`.

6. **Reorg cleanup** (same gap as #318/#452) — `INSERT OR IGNORE` keeps re-index safe but doesn't remove orphaned rows. Future hardening.

No invariant or protocol surface touched. Pure read indexer + analytics.

PR #468, opened 2026-05-07
fix(#362): auto-close zombie channels + replace fake desktop reputation with on-chain volume

### LINKED ISSUE #362: New providers see reputation as 0 despite having buyers/activity
## Bug Description

When a new provider joins the Antseed network, their displayed reputation shows as 0 even after they start receiving buyers and completing inference requests. This creates a confusing UX — a provider with active buyers still appears untrusted.

## Context

From a real user report in the Antseed community:

> It was saying my reputation is 0, but if I'm getting buyers then it seems like it's all good.

The provider was successfully routing inference requests and receiving buyers, yet the reputation score was still 0.

## Expected Behavior

New providers who are actively serving buyers should have a reputation score that reflects their actual activity, not just a static initial value of 0.

## Possible Causes

1. Reputation not initialized on-chain when a new provider first starts serving requests
2. Off-chain reputation not synced back to the on-chain registry
3. Display bug showing raw on-chain value that hasn't been seeded with initial reputation

## Additional Context

This affects the trust signal that buyers use to select providers. A new provider with 0 reputation, even with active buyers, may be disadvantaged in provider selection algorithms that weigh reputation.

kotevcode:
Closes #362.

## Why

Two coupled symptoms in #362:

1. **New providers see reputation as 0** despite serving real traffic. Buyers display `onChainChannelCount` from `AntseedChannels._agentStats[agentId].channelCount` — but that counter only ticks on `close()` / `withdraw-with-settle`, never on `reserve()` or `settle()`. Sellers in steady state reserve+settle but rarely close, so the counter never grows.
2. **Desktop "Rep" column was a lie**: `peer-cache.ts` literally hardcoded `reputation: 100` for every peer. It only affected sort order, never gated routing, and was never tied to anything on chain.

### Commit 1 — `fix(node): auto-close zombie channels so seller reputation accrues`

In `SellerPaymentManager.checkTimeouts`, when a channel is past its nominal deadline, the buyer is gone, and there is no signed spending auth, the seller now closes on-chain instead of evicting locally:

```ts
channelsClient.close(channelId, onChainSettled, "0x", "0x")
```

This works because `AntseedChannels.sol:288-294` skips `_settleSpend` (and signature verification) when `finalAmount == channel.settled`. The contract still:

- ticks `_agentStats[agentId].channelCount` → reputation accrues for new providers (the headline #362 symptom)
- decrements `activeChannelCount[seller]` → unstake stops being blocked (Gekko's secondary observation in the issue)
- releases the buyer's reserved deposit back to `AntseedDeposits` automatically

**Safety:**
- New private `_closeWithoutAuth` helper reuses the existing `_closingChannels` mutex and `MAX_CLOSE_RETRIES` retry budget.
- Reads on-chain `settled` from `getSession` first, guarding the post-restart case where local `accepted=0` but on-chain `settled>0` (would otherwise revert with `FinalAmountBelowSettled`).
- Falls back to local-only eviction when `getSession` returns `unknown` or retries are exhausted, preserving prior behavior in the bad cases.

**Tests:** +4 cases in `seller-payment-manager.test.ts` covering closes-past-deadline, respects-deadline, retry-exhaustion fallback, and "uses on-chain settled, not local 0".

### Commit 2 — `refactor(desktop): replace placeholder "reputation" with on-chain volume`

Tear out the `reputation: 100` placeholder everywhere and replace with `onChainTotalVolumeUsdcMicros` (lifetime settled USDC from `getAgentStats`, already populated into `buyer.state.json`). Mirrors what the CLI has been doing for a while via `formatUsdcVolume`.

| Surface | Before | After |
|---|---|---|
| Peers view column | "Rep" (`100` for everyone) | "Volume" (real `$X.XX`, default sort desc) |
| Overview view column | "Rep" | "Volume" |
| Session approval card | "X reputation" line | `$X.XX volume`, `N session(s)`, `Yd in network` — only the rows we have data for |
| Sort behavior | tied at `100` | volume desc, nulls sink to bottom, tiebreak `lastSeen` |

`null` vs `0` is meaningful: `null` = "no chain enrichment yet" (renders as `—`), `0` = "agent has zero settled volume" (renders as `$0`). New `formatUsdcVolume()` helper in `core/format.ts` mirrors the CLI's version.

## Verification

- Node typecheck: clean
- Node tests: **602/602** (37/37 in `seller-payment-manager.test.ts` incl. the 4 new cases)
- Desktop main + renderer typecheck: clean
- Desktop main tests: **68/68**
- Renderer Vite build: clean

## Out of scope / follow-ups

- `pi-chat-engine.ts` has a misnamed `onChainActiveChannelCount` field that actually carries the settled count. Touches tests/discover-rows; left for a focused follow-up.
- Activity-tracking signals beyond settled volume (e.g. live channel count, recent activity) are orthogonal — once auto-close ticks the on-chain counter, settled volume + channel count are the right durable signals.

### DIFF
--- a/apps/desktop/src/main/main.ts
+++ b/apps/desktop/src/main/main.ts
@@ -878,8 +878,8 @@ ipcMain.handle('payments:get-peer-info', async (_event, peerId: string) => {
       data: {
         peerId: peer.peerId,
         displayName: peer.displayName ?? null,
-        reputation: peer.reputation ?? 0,
-        onChainChannelCount: (peer as Record<string, unknown>).onChainChannelCount ?? null,
+        onChainChannelCount: peer.onChainChannelCount ?? null,
+        onChainTotalVolumeUsdcMicros: peer.onChainTotalVolumeUsdcMicros ?? null,
         onChainGhostCount: (peer as Record<string, unknown>).onChainGhostCount ?? null,
         evmAddress: peer.peerId ? peerIdToAddress(peer.peerId) : null,
         timestamp: (peer as Record<string, unknown>).timestamp ?? null,
--- a/apps/desktop/src/main/peer-cache.ts
+++ b/apps/desktop/src/main/peer-cache.ts
@@ -11,7 +11,16 @@ export type DashboardNetworkPeer = {
   inputUsdPerMillion: number;
   outputUsdPerMillion: number;
   capacityMsgPerHour: number;
-  reputation: number;
+  /**
+   * Lifetime settled volume in USDC base units (micros), from on-chain
+   * AntseedChannels.getAgentStats. `null` when chain RPC has not populated
+   * the stat (e.g. cold start, no RPC configured). The Peers/Overview tables
+   * sort and render by this; the prior "reputation" field was a static `100`
+   * placeholder and only ever sorted peers, never gated routing — see #362.
+   */
+  onChainTotalVolumeUsdcMicros: number | null;
+  /** Settled-channel count, from AntseedChannels.getAgentStats. `null` when chain RPC unavailable. */
+  onChainChannelCount: number | null;
   lastSeen: number;
   /**
    * Last successful transport-level contact with the peer. This can be fresher
@@ -129,7 +138,12 @@ export function parsePeerFromRaw(pr: Record<string, unknown>): DashboardNetworkP
     inputUsdPerMillion: Number(pr.defaultInputUsdPerMillion) || 0,
     outputUsdPerMillion: Number(pr.defaultOutputUsdPerMillion) || 0,
     capacityMsgPerHour: (Number(pr.maxConcurrency) || 0) * 60,
-    reputation: 100,
+    onChainTotalVolumeUsdcMicros: typeof pr.onChainTotalVolumeUsdcMicros === 'number'
+      ? pr.onChainTotalVolumeUsdcMicros
+      : null,
+    onChainChannelCount: typeof pr.onChainChannelCount === 'number'
+      ? pr.onChainChannelCount
+      : null,
     lastSeen: Number(pr.lastSeen) || Date.now(),
     lastReachedAt: Number(pr.lastReachedAt) || null,
     source: 'dht',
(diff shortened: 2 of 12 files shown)


kotevcode:
@claude

Augustas11:
Nice fix — the contract path and the desktop swap both check out, and the four new tests pin the load-bearing behaviors (especially the post-restart `settled=250_000n` regression guard at `seller-payment-manager.test.ts:920`).

Three small things before merge:

1. **`chat.ts` auto-retry plumbing removal isn't mentioned in the PR body.** The diff removes `PAYMENT_AUTO_RETRY_DELAY_MS`, `chatRetryTimers`, `clearPaymentRetry`, `scheduleChatRetry`, `streamCompletedAtByConversation`, and the `retryCtx` parameter from `handlePaymentRequired`. If this is intentional dead-code cleanup (which it looks like), a sentence in the PR description would help reviewers understand the scope. Otherwise it could be split into its own commit.

2. **`ConfigFormData.minRep` is now dead** at `apps/desktop/src/renderer/core/state.ts:52`. Worth deleting in this PR while you're touching the surrounding state shape — leaving it makes future "where does minRep get used?" searches misleading.

3. **Consider one more regression test** — `_closeWithoutAuth` is gated on `accepted === 0n`, where `accepted = this._acceptedCumulative.get(channel.sessionId)`. After process restart, is `_acceptedCumulative` hydrated from the store before `checkTimeouts` first runs? If a restart leaves a channel where the on-chain `settled > 0` *and* the seller has the matching signed auth in the channel store but `_acceptedCumulative` is still 0, this code path could incorrectly zombie-close instead of routing to `settleSession`. A test seeding `latestSpendingAuthSig` + non-zero `previousConsumption` and asserting `settleSession` is called (not `_closeWithoutAuth`) would lock that in.

For invariant impact: `close(channelId, settled, "0x", "0x")` with `finalAmount == channel.settled` correctly skips `_settleSpend` per `AntseedChannels.sol:291`, so no signature is verified — INV-06/07/56 are not touched. INV-27 actually improves because honest sellers can now drain `activeChannelCount` and eventually unstake. No new replay surface introduced.

Inline test comment would also help: at `seller-payment-manager.test.ts:856`, a one-liner explaining why `mgr.onBuyerDisconnect(buyerIdentity.peerId)` is called after constructor (because the constructor hydrates `_activeBuyers` from the store) would save the next reader some squinting.

kotevcode:
I actually changed too many things since and this needs a complete rewrite

kotevcode:
Closing this in favor of smaller focused PRs. The node zombie-channel cleanup is now split out in #554; the reputation display/setup follow-up will be opened separately.

PR #495, opened 2026-05-10
Fix/workspace restore on conv switch

Boris-:
## What

Fixes workspace restoration when switching between chat conversations in AntStation.

- Persists each conversation’s workspace path.
- Restores the saved workspace when opening a conversation.
- Updates the chat system prompt to reflect the per-chat workspace model.
- Hardens workspace switching against async races during rapid conversation switches.
- Ensures sends use the conversation/session workspace instead of relying only on mutable global workspace state.

## Why

Chats can be tied to different folders/repos, but switching conversations previously did not reliably restore the correct workspace. This could leave the UI showing or using the wrong project context, especially when moving between conversations from different workspaces.

The additional hardening prevents stale async workspace restore calls from overwriting the latest selected workspace, and ensures chat sends execute in the workspace associated with the conversation.

## Testing

Added/updated renderer chat routing tests for workspace behavior:

- Opening a conversation restores its saved workspace path.
- Opening a conversation does not call workspace switching when the saved workspace already matches the current workspace.
- Rapid conversation switching keeps the latest conversation’s workspace selected and re-applies it if a stale async restore completes later.

Validated with:

- `pnpm -C apps/desktop typecheck:renderer`
- `pnpm -C apps/desktop exec tsc -p tsconfig.main.json --noEmit`
- `pnpm -C apps/desktop exec tsc --noEmit --target ES2022 --module ESNext --moduleResolution Bundler --jsx react-jsx --types node,vite/client src/renderer/modules/chat.peer-routing.test.ts`
- `pnpm -C apps/desktop exec tsx --test src/renderer/modules/chat.peer-routing.test.ts`
- `git diff --check`

## Checklist

- [ ] `pnpm run build` passes
- [ ] `pnpm run test` passes
- [x] Breaking changes documented

### DIFF
--- a/apps/desktop/src/main/chat-system-prompt.ts
+++ b/apps/desktop/src/main/chat-system-prompt.ts
@@ -65,7 +65,9 @@ export function buildAntstationSystemPrompt(
   return `${resolvedBasePrompt}
 
 Workspace model:
-- All chats share the selected workspace path at the app level.
-- Conversation history is per chat, but repo and directory context are shared across chats until the workspace changes.
+- Each chat has its own workspace (folder/repo) that is stored when the chat is created.
+- When you switch to a different chat, the workspace automatically switches to that chat's stored workspace.
+- When starting a new chat, it uses whatever workspace is currently selected.
+- The workspace indicator in the UI shows the current workspace for the active chat.
 `.trim();
 }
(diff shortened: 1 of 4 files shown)


kotevcode:
@claude

Boris-:
Addressed @claude review feedback and pushed e8d1e58c.

Changes:

Guarded stale chooseWorkspace() failure reset with token === workspaceSwitchToken.
Reset desiredWorkspacePath on failed/thrown restoreWorkspace() only when it is still the active token.
Snapshotted workspace path at openConversation() start before async conversation load.
Added comment documenting SessionManager.getCwd() persisted cwd dependency.
Added regression test for stale manual workspace failure not overwriting the latest desired conversation workspace.

PR #508, opened 2026-05-12
Design assistant chat activity separation

SheferKagan:
## Summary

Adds the first production pass for inline assistant activity in the chat flow.

This builds on the assistant turn activity separation foundation branch and keeps the activity inside the assistant message rather than introducing a sidebar. The goal is a compact, Codex-style hierarchy where the final answer stays readable and the process/tool details are available when the user wants them.

## What changed

- Adds a collapsible `Worked` / `Working` section inside assistant turns.
- Keeps normal assistant text visible in the main response flow.
- Groups tool activity into compact summary rows, such as explored files or ran commands.
- Adds nested expansion for individual tool details and outputs.
- Keeps the work section collapsed by default so the answer stays primary.
- Uses the foundation branch’s final-response copy behavior so copied answers skip progress/update lines.

## Notes

The UI styling here is proposed and does not have to be final exactly as-is. The main goal of this PR is to establish the inline hierarchy and interaction model for assistant activity.

## Validation

- `pnpm --filter=@antseed/desktop exec tsc -p tsconfig.renderer.json --noEmit`
- `git diff --check`

### DIFF
--- a/apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss
+++ b/apps/desktop/src/renderer/ui/components/chat/ChatBubble.module.scss
@@ -647,6 +647,32 @@
 }
 
 /* ===================================================================
+   Assistant activity timeline
+   =================================================================== */
+
+.assistant-work-header {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  margin: 4px 0 12px;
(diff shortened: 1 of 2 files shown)


kotevcode:
#534 did something simpler

PR #509, opened 2026-05-12
fix(website): resolve network page filter and rendering issues

twfhyr:
This commit fixes three interconnected bugs on the /network/ page that caused incorrect search results, stale React renders, and inflated service counts.

Issues Fixed
--------------

1. Service name search matched peer displayName
   - The search filter checked `m.id` which was composed as `${service}::${peer.displayName}`.
   - Searching for "Open Forge" would match every service from that peer because "Open Forge" was part of `m.id`.
   - Fix: Search now explicitly checks `m.serviceId` and `m.name`, removing the `m.id` from the search condition.

2. React rendered stale/incorrect rows
   - `m.id` used `peer.displayName` which is NOT unique. Multiple peers share names like "Antseed Node".
   - This created duplicate React keys. React's reconciliation algorithm reuses DOM nodes by key, so stale rows from previous renders were displayed instead of the filtered set.
   - The result: the table showed wrong services while the "N services found" counter was correct.
   - Fix: Generate truly unique `id` values by appending a row index: `${service}::${peerId}::${rowIdx}`.

3. Service count in stats bar was inflated
   - `uniqueServiceCount` used `models.map(...).length` which counts total rows (including duplicates across peers).
   - Fix: Use `new Set(models.map(m => m.serviceId)).size` to count unique models only.

4. "Supports prompt caching" checkbox had no effect
   - The UI toggled the checkbox state but never applied a filter in the `models.filter()` logic.
   - Fix: Added `cachedInputPrice` to the `ServiceRow` type and filter: `if (filters.supportsCaching && m.cachedInputPrice === 0) return false;`.

5. Provider chip filters were not triggering the Filters button indicator
   - The Filters button only showed "active" for slider/checkbox changes.
   - Fix: Include `query`, `providerFilter`, and `tagFilter` in the `hasActiveFilters` check.

Verification
------------
- Type "kimi" in the search box → only kimi-related services are shown.
- Click the "antseed-zh" provider chip → exactly 2 rows are displayed.
- Check "Supports prompt caching" → services without cached input pricing are hidden.
- Stats bar "Services" shows the actual count of unique models.

## What

Brief description of the change.

## Why

Why is this change needed?

## Testing

How was this tested? What tests were added or updated?

## Checklist

- [ ] `pnpm run build` passes
- [ ] `pnpm run test` passes

### DIFF
--- a/apps/website/src/pages/network.tsx
+++ b/apps/website/src/pages/network.tsx
@@ -119,11 +119,11 @@ function buildServiceRows(peers: PeerMetadata[]): ServiceRow[] {
   // First pass: count how many peers serve each service
   const peerCountMap = new Map<string, Set<string>>();
   for (const peer of peers) {
-    const pName = peer.displayName ?? peer.peerId.slice(0, 12);
+    const pId = peer.peerId.slice(0, 12);
     for (const ann of peer.providers) {
       for (const service of ann.services) {
         if (!peerCountMap.has(service)) peerCountMap.set(service, new Set());
-        peerCountMap.get(service)!.add(pName);
+        peerCountMap.get(service)!.add(pId);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


kotevcode:
Thanks for the contribution!

PR #512, opened 2026-05-13
feat(metadata): v2 format with service/model hash (word[4])

Augustas11:
## Problem

The `ChannelSettled` event carries no model/service information. The metadata field (`bytes`) currently ABI-encodes 4 × uint256:

```
[version=1, inputTokens, outputTokens, requestCount]
```

This makes it impossible for explorer/indexer tooling to answer "which model was used?" for a given settled channel, since model selection is an off-chain routing decision.

## Solution — Metadata v2

Add a fifth word: `serviceHash = uint256(keccak256(utf8(serviceName)))`, with `version = 2` as a discriminant. The encoding stays backward-compatible: v1 decoders that read 4 words continue to work; v2 decoders check `word[0]` and conditionally read `word[4]`.

```
v1 (4 words): [1, inputTokens, outputTokens, requestCount]
v2 (5 words): [2, inputTokens, outputTokens, requestCount, keccak256(serviceName)]
```

Zero serviceHash means "unknown" (backward-compat with callers that omit it).

### `packages/node/src/payments/evm/signatures.ts`
- Add `serviceHash?: bigint` to `SpendingAuthMetadata`
- Add `hashServiceName(name: string): bigint` helper — `uint256(keccak256(toUtf8Bytes(name)))`
- `encodeMetadata`: emits v2 (5 words) when `serviceHash` is present and non-zero; falls back to v1 (4 words) otherwise
- Add `decodeMetadata` helper for the explorer to consume
- Keep `METADATA_VERSION` alias for backward-compat consumers

### `packages/node/src/payments/buyer-payment-manager.ts`
- Add `_currentService: Map<string, string>` (sellerPeerId → latest service slug)
- `signPerRequestAuth`: when `responseStats.service` is set, update `_currentService` and include `serviceHash` in metadata
- `handleNeedAuth`: when `buyerService` is resolved from `_requestService`, update `_currentService` and include `serviceHash` in metadata
- `cleanupSession`: clear `_currentService` on session teardown

### `packages/contracts/AntseedChannels.sol`
- NatSpec on `ChannelSettled` event documenting v2 layout — no ABI or logic changes (contract already accepts arbitrary `bytes calldata metadata` and only verifies `keccak256(metadata)`)

### `packages/contracts/test/AntseedChannels.t.sol`
- `test_settle_v2MetadataEmitsServiceHash`: reserve a channel, settle with v2 metadata carrying `keccak256("gpt-4o")`, assert `ChannelSettled` emits the full v2 bytes, assert decoded fields match

## Why no contract code change?

The `AntseedChannels` contract treats `metadata` as an opaque `bytes calldata` blob. It hashes it for EIP-712 verification and emits the raw bytes — no decoding. The 5-word v2 payload hashes correctly without any Solidity change.

## Migration path

- **New buyers** using updated `@antseed/node` will emit v2 metadata on every signed auth once a service slug is known for that channel
- **Old buyers** (v1) continue to work — contract is unchanged, explorers/indexers that read 4 words are unaffected
- **Explorer backfill**: impossible for historical events (no model in v1 data); only new post-deploy events will carry `serviceHash`

## Test plan
- [ ] `forge test --match-test test_settle_v2MetadataEmitsServiceHash` passes
- [ ] Existing `AntseedChannels.t.sol` tests unaffected (v1 encoding path unchanged)
- [ ] TypeScript: `encodeMetadata({ ..., serviceHash: hashServiceName('my-model') })` produces 5-word ABI encoding
- [ ] TypeScript: `encodeMetadata({ ... })` (no serviceHash) produces identical 4-word v1 output as before

### DIFF
--- a/packages/contracts/AntseedChannels.sol
+++ b/packages/contracts/AntseedChannels.sol
@@ -80,6 +80,11 @@ contract AntseedChannels is EIP712, Pausable, Ownable, ReentrancyGuard {
 
     // ─── Events ─────────────────────────────────────────────────────
     event Reserved(bytes32 indexed channelId, address indexed buyer, address indexed seller, uint128 maxAmount);
+    /// @dev `metadata` is ABI-encoded as N × uint256 words.
+    ///      v1 (4 words): [version=1, inputTokens, outputTokens, requestCount]
+    ///      v2 (5 words): [version=2, inputTokens, outputTokens, requestCount, serviceHash]
+    ///      where serviceHash = uint256(keccak256(bytes(serviceName))), zero = unknown.
+    ///      The contract does not decode metadata; it only verifies keccak256(metadata).
     event ChannelSettled(bytes32 indexed channelId, address indexed buyer, address indexed seller, uint128 cumulativeAmount, uint128 delta, uint128 totalSettled, uint256 platformFee, bytes metadata);
     event ChannelClosed(bytes32 indexed channelId, address indexed buyer, address indexed seller, uint128 settledAmount, uint128 refund);
     event ChannelTopUp(bytes32 indexed channelId, address indexed buyer, address indexed seller, uint128 additionalAmount, uint128 newDeposit);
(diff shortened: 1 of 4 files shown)


kotevcode:
a payment channel can contain multiple service though

PR #517, opened 2026-05-13
Docs/emissions v2 update

twfhyr:
## What

Update payment protocol documentation to reflect the current `AntseedEmissionsV2` contract deployed at migration epoch 4, replacing the legacy `AntseedEmissions` (now depleted).

### Changes by file

**`docs/protocol/spec/04-payments.md`**
- Rewrote Section 8 (Emission Distribution) to document `AntseedEmissionsV2` mechanics:
  - V2 backward compatibility with V1 points through epoch 4
  - Automatic epoch advancement via block timestamp
  - Emission schedule: `e_e = e_0 / 2^(epoch / HALVING_INTERVAL)`
  - Corrected split: **Seller 50% / Buyers 20% / Reserve 15% / Team 15%** (fixes stale 65/25/10/0 table)
  - Points accrual via `accrueSellerPoints` / `accrueBuyerPoints` / `accruePoints`
  - Optional `IAntseedPointsPolicy` hook
  - Per-epoch pro-rata distribution and seller/buyer caps (`maxSellerSharePct`, `maxBuyerSharePct`)
  - Seller claiming: locked rewards pool (`AntseedSellerRewardsPool`) with `AntseedSellerUnlockPolicy`
  - Buyer claiming via operator (`claimBuyerEmissions`)
  - Reserve & Team flush functions (`flushReserve`, `flushTeam`)
- Updated Section 10 (Contract Architecture) to include new V2 contracts
- Updated interaction flow references to V2

**`docs/protocol/diem-proxy.md`**
- Updated architecture diagram and deployment runbook references from `AntseedEmissions` to `AntseedEmissionsV2`

**`apps/website/docs/protocol/payments.md`**
- Updated Smart Contracts table to list V2, rewards pool, unlock policy, and legacy V1
- Added Emissions subsection:
  - V2 replacement at epoch 4 with backward compatibility
  - Emission split (50/20/15/15)
  - Seller/buyer claim mechanics

## Why

The protocol spec and website docs still described the legacy `AntseedEmissions` V1 contract (65/25/10 split, gas-accumulator claim, `advanceEpoch`, etc.), which was replaced by `AntseedEmissionsV2` after migration epoch 4. The V2 contract has different mechanics (locked rewards pool, unlock policy, pro-rata distribution, different emission split). The old docs were misleading for developers reading the contract source and users checking the website.

## Testing

- Verified `pnpm run build` passes in the `apps/website` workspace
- Verified `AntseedEmissionsV2.sol` source matches documented behavior (automated epochs, legacy combine logic, locked pool claiming)
- Double-checked contract address and migration epoch number against on-chain data

## Checklist

- [x] `pnpm run build` passes
- [x] `pnpm run test` passes

### DIFF
--- a/apps/website/docs/protocol/payments.md
+++ b/apps/website/docs/protocol/payments.md
@@ -128,15 +128,26 @@ USDC on Base. 6 decimal places. All on-chain amounts are in USDC atomic units (1
 ## Smart Contracts
 
 ```text title="contract architecture"
-AntseedRegistry       Central address book for all protocol contracts
-AntseedStaking        Seller staking (holds stake USDC, binds to ERC-8004 agentId)
-AntseedDeposits       Buyer deposits, seller payouts (holds buyer USDC)
-AntseedChannels       Payment channel lifecycle (swappable, holds NO USDC)
-AntseedEmissions      ANTS token emissions (USDC volume-based incentives)
-ANTSToken             ANTS ERC-20 token (1.04B max supply)
+AntseedRegistry            Central address book for all protocol contracts
+AntseedStaking             Seller staking (holds stake USDC, binds to ERC-8004 agentId)
+AntseedDeposits            Buyer deposits, seller payouts (holds buyer USDC)
+AntseedChannels            Payment channel lifecycle (swappable, holds NO USDC)
+AntseedEmissionsV2         ANTS token emissions (backward-compatible replacement)
+AntseedSellerRewardsPool   Locked ANTS reserves for sellers pending unlock
+AntseedSellerUnlockPolicy  On-chain policy controlling when sellers can claim
+AntseedEmissions (legacy)  Original emissions contract — depleted since epoch 4
+ANTSToken                  ANTS ERC-20 token (1.04B max supply)
 ```
 
-Network fees are set to 4% of settlement and flow to the Protocol Reserve, not to a company. The Protocol Reserve is intended to support long-term network sustainability, trust, utility, and alignment. Seller ANTS emissions are currently tracked but locked in a dedicated Provider Pool pending stronger validation; future claims may be subject to verification or slashing.
+Network fees are set to 4% of settlement and flow to the Protocol Reserve, not to a company. The Protocol Reserve is intended to support long-term network sustainability, trust, utility, and alignment.
+
+### Emissions
+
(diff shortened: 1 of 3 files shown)


Augustas11:
Solid alignment of the spec with V2. The math, locked-pool semantics, and backward-compat description all match `AntseedEmissionsV2.sol` and the invariant set (INV-39, INV-40, INV-45, INV-46). Removing the V1 Synthetix `rewardPerPointStored` accumulator math is correct — V2 uses per-epoch pro-rata snapshots, not a continuous accumulator.

Small fixes / additions before merge:

1. **`docs/protocol/diem-proxy.md` line 34 has a one-character indent shift** in the ASCII diagram (`        │ AntseedCh-` instead of `       │ AntseedCh-`). It misaligns the box. Quick fix.

2. **Pin a concrete `MIGRATION_EPOCH` value.** The doc says "epochs `<= MIGRATION_EPOCH`" / "depleted since epoch 4" but the latter is in `payments.md` only. Saying `MIGRATION_EPOCH = 4` explicitly in `spec/04-payments.md:222–230` would let readers map to wall-clock dates.

3. **Verify `accruePoints(channelId, buyer, seller, pointsDelta)`** described at `spec/04-payments.md:179` actually exists on the deployed `AntseedEmissionsV2`. If it's only planned, mark it as "(future)" rather than current API.

4. **Document the default of `AntseedSellerUnlockPolicy.canClaimSellerUnlocked`** — per INV-46 it's "all sellers locked" by default. Sellers wondering why their first claim went to `AntseedSellerRewardsPool` and not their wallet will appreciate the heads-up.

5. **`flushTeam()` requires `registry.teamWallet()` to be set** — worth a one-liner in `spec/04-payments.md:215-220` so operators don't get a surprise revert.

6. **Cross-track:** `CLAUDE.md` (root) still says "52M max supply" — this PR is consistent with code (1.04B in `ANTSToken.sol:10`) and with `payments.md:135`. While you're touching docs, consider also fixing `CLAUDE.md` so the documentation no longer disagrees with itself by a 20× factor (INV-L6 / hotspot H6 in our local notes).

7. **Numeric defaults for the per-epoch caps** — `spec/04-payments.md:197–200` says "default historically ~15%" and "default 5%". Filling in the concrete on-chain defaults (whatever was set at deploy) would make this doc self-contained.

Nothing here is blocking — these are all polish to make the doc the canonical source of truth.

PR #550, opened 2026-05-16
fix(payments): resolve topUp race condition causing premature channel closure

kotevcode:
## Problem

Channels are being prematurely closed and undercharging buyers due to an on-chain `topUp()` race condition.

### Root Cause

The smart contract's `topUp()` requires **85% of the current deposit to be settled** before allowing a top-up (`TopUpThresholdNotMet` error `0x1ea4506b`). The buyer was sending the topUp ReserveAuth **before** the SpendingAuth that would bring the seller's accepted cumulative high enough to meet this threshold.

### Failure Sequence (from logs)

1. Seller sends NeedAuth with `required=1,081,743` (exceeds ceiling of 1,000,000)
2. Buyer sends topUp ReserveAuth first (ceiling 1M → 2M)
3. Buyer sends SpendingAuth for 1,081,743 second
4. Seller processes topUp → calls `topUp()` on-chain with `settleAmount=734,455`
5. **Contract reverts**: `734,455 < 850,000` (85% of 1M) → `TopUpThresholdNotMet`
6. SpendingAuth for 1,081,743 rejected (exceeds unchanged `reserveMax=1,000,000`)
7. Channel declared exhausted → closed at 734,455 instead of actual spend of 1,081,743

### Buyer side (`buyer-payment-manager.ts`)
- **Send SpendingAuth BEFORE topUp**: In `handleNeedAuth` and `extendCurrentSpendingAuth`, sign at the current ceiling first (giving seller enough to settle), then send topUp ReserveAuth
- **Lower proactive threshold**: 85% → 65% to trigger topUp earlier with deferred retry as safety net

### Seller side (`seller-payment-manager.ts`)
- **Graceful topUp failure**: When on-chain `topUp()` fails, defer it instead of crashing the flow; retry after the next SpendingAuth raises the settle amount
- **Pending topUp awareness**: Accept SpendingAuths within the pending topUp ceiling (not just current on-chain ceiling)

### Request handler (`seller-request-handler.ts`)
- Use `getEffectiveReserveMax()` (includes pending topUp) when checking exhaustion to prevent premature channel closure while topUp is being retried

### DIFF
--- a/packages/node/src/payments/buyer-payment-manager.ts
+++ b/packages/node/src/payments/buyer-payment-manager.ts
@@ -28,9 +28,11 @@ import { estimateCostFromBytes, computeCostUsdc, type ServicePricing } from './p
 
 /** Default tolerance: accept seller claims up to 1.4x buyer's estimate. */
 const DEFAULT_COST_TOLERANCE = 1.4;
-/** Fraction of reserve ceiling at which to signal a top-up is needed. */
-/** Must match or exceed contract's TOP_UP_SETTLED_THRESHOLD_BPS (85%). */
-const DEFAULT_TOPUP_THRESHOLD = 0.85;
+/** Fraction of reserve ceiling at which to signal a top-up is needed.
+ *  Trigger well before the contract's TOP_UP_SETTLED_THRESHOLD_BPS (85%)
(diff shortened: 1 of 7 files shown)


kotevcode:
@claude

PR #554, opened 2026-05-17
fix(payments): close zombie channels

kotevcode:
## Description

Closes seller-side zombie payment channels on-chain when the buyer has disconnected, no SpendingAuth exists, and the channel deadline has elapsed. The seller uses the current on-chain settled amount with empty metadata/signature, which matches the contract path that skips signature verification when no additional spend is being claimed.

## Release Notes

- Fixed stale seller payment channels so disconnected zero-auth sessions can be closed on-chain and release reserved funds.

## Types of Changes

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Chore (maintenance tasks, refactoring, or non-functional changes)

## Checklist

- [x] My code follows the code style of this project
- [ ] I have added the necessary documentation (if appropriate)
- [x] I have added tests (if appropriate)
- [x] Lint and unit tests pass locally with my changes

Tested with: pnpm -C packages/node test seller-payment-manager.test.ts

PR #561, opened 2026-05-18
fix(payments): Guard reserve headroom

kotevcode:
## Description

Adds a seller-side preflight reserve headroom guard before provider routing so requests are not served when their minimum/estimated max cost cannot fit inside the confirmed on-chain reserve. Also adds a recoverable `reserve_headroom_required` payment code so buyers top up the existing channel and retry, and lowers the channel top-up settled threshold from 85% to 65%.

This addresses the observed Diem case where the seller had 684445/1000000 settled/confirmed spend, a top-up failed at the previous 85% threshold, and the next request pushed cumulative spend beyond the confirmed reserve.

## Release Notes

- Sellers now request a reserve top-up before routing paid work when confirmed channel headroom is too low for the next request.
- Buyers automatically respond to recoverable reserve-headroom 402s by topping up the active channel and retrying.
- Channel top-ups are allowed once 65% of the current reserve is settled, reducing top-up failures near the reserve ceiling.

## Types of Changes

## Checklist

### DIFF
--- a/packages/contracts/AntseedChannels.sol
+++ b/packages/contracts/AntseedChannels.sol
@@ -46,7 +46,7 @@ contract AntseedChannels is EIP712, Pausable, Ownable, ReentrancyGuard {
     uint256 public PLATFORM_FEE_BPS = 200;
     uint256 public MAX_PLATFORM_FEE_BPS = 1000;
     uint256 public TIMEOUT_GRACE_PERIOD = 15 minutes;
-    uint256 public TOP_UP_SETTLED_THRESHOLD_BPS = 8500;
+    uint256 public TOP_UP_SETTLED_THRESHOLD_BPS = 6500;
 
     // ─── Enums & Structs ────────────────────────────────────────────
     enum ChannelStatus { None, Active, Settled, TimedOut }
@@ -190,7 +190,7 @@ contract AntseedChannels is EIP712, Pausable, Ownable, ReentrancyGuard {
     /**
      * @notice Top up an active channel by increasing the reserve ceiling.
      *         Seller calls this when the buyer's cumulative spending approaches
-     *         the current deposit. Requires at least 85% of the current deposit
+     *         the current deposit. Requires at least 65% of the current deposit
      *         to be settled (proven via SpendingAuth) before allowing more funds.
      *         Accepts the latest SpendingAuth to settle before raising the ceiling.
      *
@@ -223,7 +223,7 @@ contract AntseedChannels is EIP712, Pausable, Ownable, ReentrancyGuard {
             _settleSpend(channelId, channel, cumulativeAmount, metadata, spendingSig);
         }
 
-        // Require at least 85% of current deposit to be settled before topping up
+        // Require enough current deposit to be settled before topping up
         uint256 threshold = (uint256(channel.deposit) * TOP_UP_SETTLED_THRESHOLD_BPS) / 10000;
         if (channel.settled < threshold) revert TopUpThresholdNotMet();
 
--- a/packages/contracts/test/AntseedChannels.t.sol
+++ b/packages/contracts/test/AntseedChannels.t.sol
@@ -728,6 +728,10 @@ contract AntseedChannelsTest is Test {
         assertEq(channels.PLATFORM_FEE_BPS(), 300);
     }
 
+    function test_defaultTopUpSettledThresholdBps() public view {
+        assertEq(channels.TOP_UP_SETTLED_THRESHOLD_BPS(), 6500);
+    }
+
     function test_setPlatformFeeBps_revert_aboveMax() public {
         vm.expectRevert(AntseedChannels.InvalidFee.selector);
         channels.setPlatformFeeBps(1001);
@@ -805,8 +809,8 @@ contract AntseedChannelsTest is Test {
         bytes32 salt = keccak256("session-topup");
         bytes32 channelId = doReserve(salt, USDC_100, USDC_150);
 
-        // Top up with inline settle of 85 USDC (85% threshold met)
-        uint128 settleAmount = 85_000_000;
+        // Top up with inline settle of 65 USDC (65% threshold met)
+        uint128 settleAmount = 65_000_000;
         bytes memory spendingSig = signSpendingAuth(BUYER_PK, channelId, settleAmount, 5000, 2000);
 
         uint128 newMax = USDC_150;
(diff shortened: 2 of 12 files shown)


kc-zero-lab:
Reviewed as a coding agent. I think there’s a correctness blocker in the new `reserve_headroom_required` flow.

The seller can now emit `reserve_headroom_required` whenever the *forward-looking* required headroom exceeds remaining reserve, even when the accepted/settled cumulative is still far below the contract’s top-up gate. Example from the new tests: `spent=accepted=100_000`, `reserveMax=1_000_000`, large max-token estimate requires `requiredCumulativeAmount=1_100_000`. The buyer handles this by calling `topUpReserve()` directly, but `topUpReserve()` sends the current cumulative amount as-is. At 100k/1M, the seller’s on-chain `topUp()` should still fail `TopUpThresholdNotMet` because the contract now requires 65% of the current deposit to be settleable before top-up.

That creates a deadlock for large early requests:

1. seller refuses to route because there is not enough confirmed headroom,
2. buyer sends a top-up auth with only the current low cumulative,
3. seller cannot apply the top-up on-chain because the 65% threshold is not met,
4. retry gets the same `reserve_headroom_required` again.

This is different from the existing `handleNeedAuth()` top-up path, which first advances the SpendingAuth up to the old ceiling when top-up is needed. The new preflight path intentionally avoids signing undelivered spend, so it cannot satisfy the top-up threshold when `accepted < 65% * reserveMax`.

I’d suggest one of these fixes before merge:

- only emit `reserve_headroom_required` when the accepted/settleable cumulative is already above the contract top-up threshold; otherwise return a distinct non-retryable/renegotiate response, or route only if the request fits current confirmed headroom;
- or have the buyer retire/renegotiate a fresh larger reserve when headroom is needed but the current channel is below the top-up threshold;
- or change the protocol payload so the seller can explicitly distinguish “top-up is currently possible” from “this request cannot fit current headroom yet”.

Also worth adding a buyer+seller integration test for the low-spend large-request case (`accepted < 65% reserveMax` and `requiredHeadroom > remainingHeadroom`) to prove it does not retry-loop.

I tried to run the targeted node tests, but this checkout is missing `node_modules`/vitest/tsc, so I could only do static review here.

kotevcode:
Addressed the low-spend large-request retry-loop blocker.

The seller now only emits recoverable `reserve_headroom_required` when the accepted cumulative is already at the 65% top-up threshold. If confirmed headroom is insufficient before that threshold, it returns `channel_exhausted`, closes the current channel through the existing close path, and lets the buyer renegotiate a fresh reserve instead of retrying an impossible top-up.

Added coverage for the low-spend/max-token case and re-ran:
- `pnpm --filter @antseed/node test -- node-payment-pricing.test.ts buyer-payment-negotiator.test.ts payment-codec.test.ts`
- `git diff --check`

Augustas11:
## Review verdict

**APPROVE_WITH_NITS.** Headroom math is correct, the deadlock blocker @kc-zero-lab flagged is properly resolved (seller now returns `channel_exhausted` + closes when `accepted < 65%` so the buyer never retries an impossible top-up), the 65% threshold is consistent on-chain and off-chain, and tests cover all three preflight branches (top-up, close, fits). Findings below are non-blocking.

### Findings

**P2 — Worth addressing**

1. **Threshold drift risk: hardcoded 6500 BPS in two new places** — `packages/node/src/seller-request-handler.ts:633-636` (`_canTopUpAtCurrentAccepted`) and `packages/node/src/payments/buyer-payment-manager.ts:32-34` bake `6500`/`0.65` as literals. `AntseedChannels.TOP_UP_SETTLED_THRESHOLD_BPS` is owner-settable via `setTopUpSettledThresholdBps` (`AntseedChannels.sol:504`). If an operator ever changes it, the seller's preflight mis-classifies `canTopUpNow`, emits `reserve_headroom_required` when the on-chain gate will revert, and the buyer's `topUpReserve()` round-trip burns seller gas on a guaranteed `TopUpThresholdNotMet`. Suggest reading the threshold from `ChannelsClient` at startup and threading it into `seller-payment-manager`.
2. **Per-request top-up amortization** — at 65% the preflight (estimate vs remaining headroom) can fire for cheap requests with large `max_tokens` even when actual delivered cost is small. There is no per-channel minimum top-up batching, so bursty requests can churn one on-chain `topUp()` per request. Suggest `requiredHeadroom > max(remainingHeadroom, minTopUpDelta)` so micro-shortfalls fall through to the existing `NeedAuth` path.

**P3 — Minor**

1. **Parallel-request race in `topUpReserve`** — `buyer-payment-manager.ts:974-1014` reads `prevCeiling`, computes `newCeiling`, signs, sends, then commits. Two concurrent `reserve_headroom_required` 402s (pipelined requests) can both observe the same `prevCeiling` and emit duplicate ReserveAuths — the seller's second `topUp()` reverts `TopUpAmountTooLow` (INV-10). Not unsafe, but a per-seller mutex around `topUpReserve` would suppress the spurious revert/gas.
2. **Buyer skips deposit-headroom check before signing top-up** — `buyer-payment-negotiator.ts:347-359` only checks `balance.available <= 0n` before calling `topUpReserve`. PR #563 adds a stricter `available < maxReserveAmountUsdc` short-circuit on the negotiation path. Once #563 lands, mirror that check in the new `reserveHeadroomRequired` branch so the buyer doesn't sign a top-up the seller can't fund.
3. **Memory drift** — INV-09 and INV-60 still state "85%"; post-merge bookkeeping for our local invariants doc.

### Notes

- The `requiredCumulativeAmount` semantics split in `types/protocol.ts:89-104` is documented well — forward-looking headroom target (recoverable) vs claimable SpendingAuth (exhausted).
- The 100k/1M low-spend / max-token regression test (`node-payment-pricing.test.ts:557-595`) directly captures @kc-zero-lab's blocker scenario — nice coverage.
- CI green: Test + Typecheck + E2E Payment Flow all SUCCESS.
- Cross-cut with PR #554 (zombie-close): the preflight close path feeds the same `settleSession` flow #554 hardened; no obvious interaction issue.

Gekko-dot-ETH:
**Edit — redacting this:** I got the version wrong. We weren't on 0.2.87
in prod — our seller was on `@antseed/cli@0.1.121`, which pins
`@antseed/node@0.2.86`, so it resolved the CLI's nested 0.2.86 copy rather
than the 0.2.87 we had at top-level. The `0xa97ce07e…` overdraft was
0.2.86's `effectiveMax = max(reserveMax, pendingTopUp.newMaxAmount)`, which
0.2.87 already removed — so it doesn't reproduce there.
`@antseed/cli@0.1.122` already bumps the pin to 0.2.87; this was just a
stale CLI on our side. Sorry for the noise.

kotevcode:
thanks, please share the full log. can find me on tg

PR #563, opened 2026-05-18
fix(node): Stop short deposit auth retries

kotevcode:
## Description

Prevents buyers from repeatedly signing payment authorizations when their available deposit is below the reserve amount the seller will lock on-chain. The buyer now returns a normalized `payment_required` response before reserve negotiation and skips reserve top-up signatures when the added reserve cannot be covered.

## Release Notes

- Fixed repeated payment authorization retries when buyer deposits are too low to open or top up a channel
- Buyers now receive a clear `insufficient_deposits` payment response instead of hammering sellers with failing reserve attempts

## Types of Changes

## Checklist

### DIFF
--- a/packages/node/src/payments/buyer-payment-manager.ts
+++ b/packages/node/src/payments/buyer-payment-manager.ts
@@ -973,7 +973,25 @@ export class BuyerPaymentManager {
 
     const prevCeiling = this._getCeiling(sellerPeerId);
     const newCeiling = prevCeiling + this._config.maxReserveAmountUsdc;
+    const additionalReserve = newCeiling - prevCeiling;
     const deadline = Math.floor(Date.now() / 1000) + this._config.defaultAuthDurationSecs;
+
+    try {
+      const balance = await this.getBalance();
+      if (balance.available < additionalReserve) {
+        throw new Error(
+          `Insufficient buyer deposits for reserve top-up: available=${balance.available} required=${additionalReserve}`,
+        );
+      }
+    } catch (err) {
+      if (err instanceof Error && err.message.startsWith('Insufficient buyer deposits')) {
+        throw err;
+      }
+      debugWarn(
+        `[BuyerPayment] topUpReserve: unable to verify buyer deposits before signing top-up: ` +
+        `${err instanceof Error ? err.message : err}`,
+      );
+    }
 
     debugLog(`[BuyerPayment] topUpReserve: channel=${session.sessionId.slice(0, 18)}... ceiling ${prevCeiling} → ${newCeiling}`);
 
--- a/packages/node/src/payments/buyer-payment-negotiator.ts
+++ b/packages/node/src/payments/buyer-payment-negotiator.ts
@@ -42,6 +42,16 @@ interface LastResponseCost {
   service?: string;
 }
 
+class InsufficientBuyerDepositsError extends Error {
+  constructor(
+    readonly available: bigint,
+    readonly required: bigint,
+  ) {
+    super(`Buyer deposits available=${available} are below required reserve=${required}`);
+    this.name = 'InsufficientBuyerDepositsError';
+  }
+}
+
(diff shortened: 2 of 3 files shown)


Augustas11:
## Review verdict

**APPROVE_WITH_NITS.** The core fix is sound: the buyer performs an on-chain `getBuyerBalance` lookup BEFORE signing the ReserveAuth on the 402 path (`buyer-payment-negotiator.ts:411-422`) and BEFORE signing a top-up ReserveAuth (`buyer-payment-manager.ts:979-994`). Both call sites short-circuit with an `insufficient_deposits` signal, killing the wasteful sign → seller-reverts → resign loop described in the PR body. Tests cover the happy-path 402 case and the top-up case (`payment-negotiation.test.ts:565-622`). Remaining issues are wire-level concerns about how the `insufficient_deposits` code propagates to consumers and how the new `throw` interacts with the headroom branch added in #561.

### Findings

**P2 — Worth addressing**
1. Desktop UI does not differentiate `insufficient_deposits` from a normal `payment_required`. The body at `buyer-payment-negotiator.ts:288-295` carries `code: 'insufficient_deposits'` and a CLI hint in `message`, but `apps/desktop/src/main/pi-chat-engine.ts:1990-2031` routes all 402 bodies through `normalizePaymentBody` + `emitPaymentRequiredStreamError`, forwarding only `payment_required:<suggestedAmount>`. Neither `code` nor `message` reaches the renderer — the user sees the generic "approve payment" CTA, not a "top up your deposits" prompt. Recommend either a distinct `ChatStreamStopReason.kind = 'insufficient_deposits'` or attaching `body.code`/`body.message` to the existing stop reason.
2. With #561 merged, the new `reserve_headroom_required` branch in `handle402` calls `await this._bpm.topUpReserve(peer.peerId, pmux)` with no try/catch. PR #563 now makes that call THROW when deposits are short — the throw will reject `handle402` and surface as an unhandled negotiator error instead of a normalized `insufficient_deposits` body. Recommend #561's headroom branch wrap `topUpReserve` and map `'Insufficient buyer deposits'` to `returnPaymentRequired('insufficient_deposits', ...)`. Worth flagging on both PRs.
3. Test gap: no coverage for the `requestedReserveAmount` clamp at `:264` (suggestedAmount > `maxReserveAmountUsdc`), nor for the case where the prior initial SpendingAuth is preserved as a valid channel after a top-up rejection. Worth adding before relying on this guard.

**P3 — Minor**
1. `buyer-payment-manager.ts:986-994` — the rethrow guard sniffs the literal prefix `'Insufficient buyer deposits'`. If the message is ever rephrased the throw silently downgrades to `debugWarn` and signing proceeds. Use a typed sentinel (`class InsufficientDepositsError extends Error {}`) instead.
2. `requestedReserveAmount` derivation (`:260-265`) only uses `suggestedAmount`; for `channel_exhausted` payloads where `reserveMaxAmount` (already parsed at `:347`) is larger, the guard can pass while the eventual ReserveAuth locks more. Consider `max(suggested, reserveMaxAmount)`.
3. `bodyRequirements` (`:248-257`) and the inline reconstruction in `returnPaymentRequired` (`:277-287`) build very similar shapes — could collapse to one helper.

### #561 ↔ #563 interaction

When both PRs land: the seller may emit `reserve_headroom_required` asking the buyer to top up; #561's branch calls `topUpReserve` directly. With #563's new throw, an under-funded buyer's `topUpReserve` rejects with `'Insufficient buyer deposits…'`, NOT caught by the headroom branch — it bubbles up as an unhandled negotiator failure instead of a normalized `insufficient_deposits` payment_required body. The combined flow does not infinite-loop (the primary goal — well done), but the failure mode degrades from "loop forever" to "throw raw exception with no UI signal". The fix is a small try/catch in #561's headroom branch; flagging here so the two PRs are not merged in the wrong order without that wrapper.

PR #573, opened 2026-05-19
Add peer-scoped Agent access approvals

SheferKagan:
## Summary

Adds Agent access controls to desktop chat:

- `Ask first` gates risky tools before they run
- `Full access` keeps the existing no-approval behavior
- approvals are handled in the main process through the Pi tool-call hook
- remembered approvals are scoped to the peer and tool, not the workspace or all providers

The risky tools gated in `Ask first` are `bash`, `edit`, `write`, `start_dev_server`, and `web_fetch`. Read-only tools like `read`, `grep`, `find`, `ls`, and `open_browser_preview` continue without prompting.

For `bash`, commands are classified before prompting. Plain read-only commands (e.g. `ls`, `grep`, read-only `git`) auto-allow without a prompt, but only when the command is a single statement with no command substitution, newlines, or chaining — anything that could smuggle a second command past the first-word check is forced to an explicit prompt.

## Notes

The persistent allow rules are keyed by peer id. Allowing `bash` for one peer does not allow `bash` for another peer, and it does not allow other tool categories for the same peer. If there is no peer id, the UI only offers one-time approval/deny.

Old workspace-scoped permission data is ignored so it cannot become a broader approval after this change.

## UI

Adds an approval card with:

- Allow once
- Always allow <tool> for this peer
- Deny

Also updates the Agent access selector so it reads more like an interactive control.

## Validation

Ran from `apps/desktop`:

- `npm run build:main`
- `npm run typecheck:renderer`
- `npx tsx --test src/main/chat-tool-policy.test.ts`


Augustas11 (commented):
Reviewed `feat/permission-mode-selector` (head `f20edd9`). The peer+tool scoping design is solid and the main-process blocking gate is correctly implemented. Two substantive issues worth addressing before merge.

---

**[nit, security] `web_fetch` is ungated and can reach arbitrary external hosts**

`RISKY_TOOLS` at `chat-tool-permissions.ts:25` is `['bash', 'edit', 'write', 'start_dev_server']`. `webFetchTool` is registered at `pi-chat-engine.ts:1874-1875` but absent from that list, so no approval prompt fires for it.

`chat-web-fetch.ts:140-143` accepts any `http:`/`https:` URL; lines 158-166 call `fetch(parsedUrl, { method: 'GET', redirect: 'follow' })`. For JS-rendered pages, lines 126-128 load the URL in a hidden BrowserWindow. A malicious model could encode context it already has into a URL query string and exfiltrate it without an approval card.

Suggest adding `web_fetch` to `RISKY_TOOLS`, or constraining it to an allowlist of trusted domains. If intentionally ungated, a comment explaining the decision would help future reviewers.

---

**[nit] Concurrent gated tool calls — first approval becomes invisible**

`pendingToolApprovals` at `pi-chat-engine.ts:1555` is a `Map`, so the main process can hold multiple awaited approvals in parallel. But renderer state at `chat.ts:2699-2703` stores only one `chatToolApprovalRequest` and overwrites it on each new request.

If two gated tool calls arrive before the first is answered, the first approval card disappears from the UI while its main-process promise stays pending — blocked forever until session abort. Suggest queuing approval requests in the renderer one at a time, or explicitly blocking a second gated call until the first resolves.

---

**Notes (no action required)**

- No migration from old `workspaceAllow` grants is intentional (`chat-tool-permissions.ts:52-53`). Users will be re-prompted after update — worth a line in release notes so they're not surprised.
- `peerId`-absent path correctly degrades to allow-once only (`ToolApprovalCard.tsx:33-36`). Good defensive default.
- Blocking gate is correct: `tool_call` handler awaits `requestToolApproval()` at `pi-chat-engine.ts:1812-1838`; abort resolves pending approvals as deny at lines 1564-1570.
- `ToolApprovalCard` uses `role="group"` rather than a focus-trapping modal — not a correctness issue (the gate is in main process), but worth noting for a future a11y pass.

Overall: **approve with the two nits above addressed** (or explicitly documented as known limitations).

PR #590, opened 2026-05-20
feat(contracts): Add usage verification

kotevcode:
## Description

Adds a no-funds AntseedUsageVerification contract for service-scoped usage attestations by epoch. Wires verification protocol messages, local persistence, EVM client helpers, and buyer/seller managers so service usage can be committed and revealed alongside existing payment channels.

## Release Notes

- Added service-scoped usage verification facts for future reputation and analytics
- Added buyer/seller verification messaging and local persistence for commit-reveal attestations
- Added deployment and CLI config wiring for the usage verification contract

## Types of Changes

## Checklist

## Verification

- `pnpm --filter @antseed/node test`
- `pnpm run typecheck`
- `pnpm run build`
- `cd packages/contracts && forge build`
- `cd packages/contracts && forge test`
- `cd packages/contracts && forge test --match-contract AntseedUsageVerificationTest`

Note: attempted `vitest run --coverage`, but this workspace does not currently include `@vitest/coverage-v8`, so coverage reporting is unavailable without adding that dependency.

### DIFF
--- a/apps/cli/src/cli/commands/buyer/start.ts
+++ b/apps/cli/src/cli/commands/buyer/start.ts
@@ -274,6 +274,7 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
         depositsContractAddress: cryptoOverrides?.depositsContractAddress,
         channelsContractAddress: cryptoOverrides?.channelsContractAddress,
         usdcContractAddress: cryptoOverrides?.usdcContractAddress,
+        usageVerificationContractAddress: cryptoOverrides?.usageVerificationContractAddress,
       })
       let settlementEnabled = settlementEnv ?? true
 
@@ -302,6 +303,7 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
           // as `null` in buyer.state.json).
           ...(chainConfig.stakingContractAddress ? { stakingAddress: chainConfig.stakingContractAddress } : {}),
           ...(chainConfig.identityRegistryAddress ? { identityRegistryAddress: chainConfig.identityRegistryAddress } : {}),
+          ...(chainConfig.usageVerificationContractAddress ? { usageVerificationAddress: chainConfig.usageVerificationContractAddress } : {}),
           chainId: chainConfig.evmChainId,
           defaultDepositAmountUSDC: cryptoOverrides?.defaultLockAmountUSDC
             ? String(Math.round(parseFloat(cryptoOverrides.defaultLockAmountUSDC) * 1_000_000))
--- a/apps/cli/src/cli/commands/seller/start.ts
+++ b/apps/cli/src/cli/commands/seller/start.ts
@@ -419,6 +419,7 @@ export function registerSellerStartCommand(sellerCmd: Command): void {
           usdcContractAddress: config.payments.crypto?.usdcContractAddress,
           identityRegistryAddress: config.payments.crypto?.identityRegistryAddress,
           emissionsContractAddress: config.payments.crypto?.emissionsContractAddress,
+          usageVerificationContractAddress: config.payments.crypto?.usageVerificationContractAddress,
           subPoolContractAddress: config.payments.crypto?.subPoolContractAddress,
         })
         const defaultLockAmountUSDCBaseUnits = toUSDCBaseUnits(
@@ -572,6 +573,7 @@ export function registerSellerStartCommand(sellerCmd: Command): void {
             usdcAddress: paymentConfig.crypto.usdcAddress,
             identityRegistryAddress: resolveChainConfig({ chainId: paymentConfig.crypto.chainId }).identityRegistryAddress,
             stakingAddress: resolveChainConfig({ chainId: paymentConfig.crypto.chainId }).stakingContractAddress,
+            usageVerificationAddress: config.payments.crypto?.usageVerificationContractAddress ?? resolveChainConfig({ chainId: paymentConfig.crypto.chainId }).usageVerificationContractAddress,
             chainId: resolveChainConfig({ chainId: paymentConfig.crypto.chainId }).evmChainId,
           } : {}),
         },
--- a/apps/cli/src/config/types.ts
+++ b/apps/cli/src/config/types.ts
@@ -163,6 +163,8 @@ export interface PaymentsCLIConfig {
     identityRegistryAddress?: string;
     /** Deployed AntseedEmissions contract address */
     emissionsContractAddress?: string;
+    /** Deployed AntseedUsageVerification contract address */
+    usageVerificationContractAddress?: string;
     /** Deployed AntseedSubPool contract address */
     subPoolContractAddress?: string;
     /** Default lock amount per session in human-readable USDC (e.g. "1" = 1 USDC) */
--- a/packages/contracts/AntseedChannels.sol
+++ b/packages/contracts/AntseedChannels.sol
@@ -130,6 +130,15 @@ contract AntseedChannels is EIP712, Pausable, Ownable, ReentrancyGuard {
         return _agentStats[agentId];
     }
 
+    function getUsageVerificationChannel(bytes32 channelId)
+        external
+        view
+        returns (address buyer, address seller, uint256 settled)
+    {
+        Channel storage channel = channels[channelId];
+        return (channel.buyer, channel.seller, uint256(channel.settled));
+    }
+
     // ═══════════════════════════════════════════════════════════════════
     //                        CORE — RESERVE
     // ═══════════════════════════════════════════════════════════════════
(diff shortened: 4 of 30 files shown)


Augustas11 (changes requested):
Reviewed against the PR diff. Two independent reviews were obtained and only findings that **both** surfaced are included below — high-confidence intersection only.

Architecture-wise this looks like the right direction: the EIP-712 domain separation from `SpendingAuth` holds, no fund-at-risk path was found (no `payable`/`transfer`/`approve` anywhere in the new contract), and the lifecycle is genuinely decoupled from settlement so a failed `retryCommittedReveals` doesn't block payment flow. The four items below are correctness/operability concerns that I think want to land before merge.

---

## 1. Buyer signs `UsageCommit` without validating seller signature, channel ownership, or cumulative counters

**Where:** `packages/node/src/payments/usage-verification-manager.ts:235` (buyer commit handler, ~`handleCommitRequest:2135` in the file as viewed in the PR).

The buyer's pre-signing validation only checks optional service name, buyer address, and claim hash. It does not verify (a) the seller's signature on the seller-supplied claim, (b) that `claim.channelId` is a channel the buyer actually opened with this seller, (c) cumulative token counts vs the buyer's locally observed usage, (d) `paymentCumulativeAmount` vs what the buyer's payment manager has actually authorized, or (e) `expectedEpoch` freshness.

On-chain reveal only enforces `settled >= paymentCumulativeAmount`. That guarantees payment coverage but not metric truth — a malicious seller can request the buyer's signature on inflated or misattributed usage facts, and once signed, those become canonical on-chain attestations.

**Suggested fix:** add a buyer-side cross-check that validates `sellerSig`, channel ownership (`channelId` belongs to a `(buyer, seller)` pair the buyer has opened), observed cumulative usage, cumulative cost, cumulative payment, and current epoch before producing the buyer signature. Of the four items in this review, this is the one that most directly affects the soundness of the on-chain numbers.

## 2. `retryCommittedReveals` has no backoff, no attempt cap, and no terminal-vs-transient classification

**Where:** `packages/node/src/payments/usage-verification-manager.ts:175` (the retry pass), invoked from `node.ts:~1022` on every idle-settle.

`listAttestationsByStatus('committed', limit)` returns the oldest committed records every pass, regardless of why prior attempts failed. Permanent reverts (invalid nonce, non-monotonic claim, agent mismatch, persistent payment shortfall) are swallowed and the row stays `committed`, so the same hopeless attestations get re-attempted indefinitely on every idle-settle tick.

The store doesn't track `attempt_count`, `last_error`, or `next_retry_at`, so there's no way for the retry loop to back off or to demote a row to a terminal state for operator visibility.

**Suggested fix:** extend the store schema with `attempt_count`, `last_error`, and `next_retry_at`. Classify revert reasons into terminal (mark `failed_permanent`, stop retrying, surface to metrics) vs transient (exponential backoff up to a cap). The simplest cut is to gate retry on `next_retry_at <= now()` and add a permanent-failure status to the existing `status` enum.

## 3. `AntseedChannels.channels()` is decoded by struct-tuple layout — silent ABI drift risk

**Where:** `AntseedUsageVerification.sol:11` declares an inline `function channels(bytes32 channelId) ... returns (...)` mirror of the channels struct shape, and `:312` reads it to extract buyer/seller/settled.

Any reorder, add, remove, or type change to the public `Channel` struct on `AntseedChannels` will silently make verification revert or — worse — decode the wrong fields without reverting. There's no compile-time guard and no integration test against the real `AntseedChannels.sol` ABI.

**Suggested fix:** add an explicit stable accessor on `AntseedChannels`, e.g. `getUsageVerificationChannel(bytes32 channelId) external view returns (address buyer, address seller, uint256 settled)`, and have the verification contract call that instead of decoding the raw struct tuple. Pairs naturally with an integration test that pins the accessor's signature so future channel refactors break the test instead of breaking verification silently.

## 4. Zero node-side unit tests for the new manager / store / codec / mux

**Where:** PR adds three Solidity tests (`test_commitAndReveal_recordsServiceScopedDeltas`, `test_reveal_requiresPaymentCoverage`, `test_commit_rejectsWrongEpoch`) and no `*.test`/`*.spec`/`__tests__` files anywhere under `packages/node/**`.

The highest-risk surfaces in this PR — buyer validation (item 1 above), retry classification (item 2), codec rejection of malformed messages, store migration + upsert path, and the new mux routing — therefore ship without CI guard. The Foundry suite covers the contract happy path and one negative case for epoch; it doesn't cover `revealPartial` lifecycle, bad signatures, channel mismatch, restart recovery, or cross-language EIP-712 hash equivalence between the TS `computeUsageClaimHash` and the Solidity `hashUsageClaim`.

**Suggested fix:** add Vitest/Jest suites for `BuyerUsageVerificationManager` and `SellerUsageVerificationManager` (commit, reveal, retry, restart, malformed message, signature rejection), the codec encoders/decoders, the store migration/upsert path, and the mux routing. Cross-language EIP-712 equivalence is also worth a small parity test — both sides need to hash to the same `claimHash` or the buyer rejects every commit silently.

---

**Overall:** `REQUEST_CHANGES`. The architecture direction is the right one for verifiable reputation on top of the payment channel substrate; the four items above are the ones that would change my read of "merge-ready" to "merged with confidence."

Happy to take a follow-up look once items 1 and 4 land — those two together would close most of the operational risk on the off-chain lifecycle.

Augustas11 (changes requested):
Thanks for the follow-up push. Three of the four findings from the prior review are substantially addressed: buyer-side validation now verifies the seller's EIP-712 signature, epoch freshness, and channel-ownership/counter bounds before signing (F1); the retry loop has capped exponential backoff, terminal-vs-transient classification, and per-row attempt tracking (F2); and the new `usage-verification.test.ts` covers the critical happy-path and adversarial flows (F4). That is meaningful progress on a complex surface.

Three issues remain from this pass:

## 1. `getUsageVerificationChannel` not implemented in `AntseedChannels.sol`

**Where:** `packages/contracts/AntseedUsageVerification.sol:363–382` (`_readChannelForUsageVerification`) and `packages/contracts/AntseedChannels.sol` (accessor absent).

The new `_readChannelForUsageVerification` tries `IAntseedChannelsUsageAccessor.getUsageVerificationChannel` first and falls back to the legacy 9-field struct decode. The design is right — but `AntseedChannels.sol` does not implement `getUsageVerificationChannel`. On all currently deployed contracts the `try` always reverts (selector not found), so execution always falls through to the legacy tuple decode. The ABI-drift risk the commit set out to fix is deferred rather than closed: any reorder/add/remove of `Channel` fields in a future `AntseedChannels` redeploy will still silently mis-decode.

**Suggested fix:** Add `function getUsageVerificationChannel(bytes32 channelId) external view returns (address buyer, address seller, uint256 settled)` to `AntseedChannels.sol`. A simple view that reads `channels[channelId]` and returns the three fields — no logic change, no gas impact on existing paths. Once that lands, the primary path in `_readChannelForUsageVerification` fires and the legacy fallback becomes a true backwards-compatibility shim for pre-upgrade deployments.

## 2. `_validateAgainstBuyerState` bypassed when `sellerPeerId` is absent

**Where:** `packages/node/src/payments/usage-verification-manager.ts:416`.

`_validateAgainstBuyerState` returns `null` (accept) when `!this._validationSource || !sellerPeerId`. The production wiring passes `sellerPeerId` correctly today, but `trackRequestService`'s `sellerPeerId` argument is optional, so any future call path that omits it silently skips the channel-ownership and cumulative-counter checks while still passing signature/epoch validation. The strongest buyer-side guard becomes silently inert.

**Suggested fix:** Either make `sellerPeerId` required in `trackRequestService`, or add a log/assert when `sellerPeerId` is absent and the validation source is set (the combination "we have a source but no peer id" is likely a bug at the call site). A test that explicitly covers the bypass path would also help.

## 3. No reveal-window deadline on `revealPair` / `revealPartial`

**Where:** `packages/contracts/AntseedUsageVerification.sol` — `revealPair` (line ~264) and `revealPartial` (line ~301); `_validateReveal` enforces no epoch/timestamp deadline.

Both reveal functions accept a valid `commitHash` at any point after the commit, including epochs arbitrarily far in the past. A late reveal (e.g., a node crashing for weeks then coming back) permanently increments the epoch's usage counters retroactively. Over multiple epochs this creates stats records that no longer correspond to when the usage actually occurred. Downstream analytics, reward calculations, and reputation systems reading the on-chain stats would see silently back-dated usage.

This was flagged in the original review and was not addressed in this push — noting it again since the stats-integrity concern is proportional to the protocol's maturity.

**Suggested fix:** Enforce a maximum reveal lag (e.g., `require(currentEpoch() <= commit.epoch + REVEAL_WINDOW, RevealWindowExpired())` with `REVEAL_WINDOW` a governance-settable constant). Unrevealable commits after the window could be swept into a `lapsed` bucket counted separately from real usage. Exact window length is a protocol design choice — any finite bound closes the retroactive-mutation path.

---

Happy to re-review once these are addressed. The overall direction is correct and the buyer-validation + retry work especially is a solid improvement.

PR #616, opened 2026-05-27
fix(metering): zero-cost receipts for failed requests

### LINKED ISSUE #614: usage_receipts written with cost > 0 for requests that returned 4xx/5xx
Noticed while reading `dist/metering/seller-session-tracker.js` in `@antseed/node 0.2.88`.

`recordMetering()` receives a `statusCode` and creates a `MeteringEvent` that carries it through. It then unconditionally generates and inserts a `UsageReceipt` priced from the token estimates, regardless of whether the response succeeded:

```js
// seller-session-tracker.js — recordMetering() (excerpt)
metering.insertEvent(event);  // event has statusCode
if (this._receiptGenerator) {
  const estimatedCostUsd = (
    tokens.inputTokens  * pricing.inputUsdPerMillion  +
    tokens.outputTokens * pricing.outputUsdPerMillion
  ) / 1_000_000;
  const unitPriceCentsPerThousandTokens =
    Math.max(0, (estimatedCostUsd / tokens.totalTokens) * 1000 * 100);
  const receipt = this._receiptGenerator.generate(
    session.sessionId, event.eventId, providerName,
    buyerPeerId, event.tokens, unitPriceCentsPerThousandTokens
  );
  metering.insertReceipt(receipt);              // cost_cents > 0 always
  session.totalCostCents += receipt.costCents;  // session totals inflated
}
```

The handler in `seller-request-handler.js` (around line 320 in 0.2.88) calls `recordMetering()` from the response path after errors too: the provider-exception catch block assigns `statusCode = 500` and then falls through to the same metering call. The same path is reached when the provider call short-circuits with 402.

Impact: anyone treating `usage_receipts.cost_cents` as the seller's earned-revenue ledger (or `sessions.total_cost_cents`, which is incremented from the same receipt) sees inflated totals from 4xx/5xx responses that were never delivered. Buyers who are protected from being billed for failed work would expect the receipt ledger to mirror that contract.

Suggested fix direction: gate the receipt-insert block on `statusCode >= 200 && statusCode < 400`, or persist the receipt with `costCents = 0` when the request failed so the audit trail survives without inflating revenue numbers.

ilonagaja509-glitch:
Fixes #614

## Problem
`recordMetering()` generates `UsageReceipt` with `costCents > 0` even for failed requests (4xx/5xx). This inflates seller revenue and charges buyers for failed work.

## Solution
Gate receipt cost calculation on `statusCode >= 200 && statusCode < 400`. Failed requests get zero-cost receipts (preserving audit trail).

## Changes
- Added `isSuccess` check based on statusCode
- `estimatedCostUsd` is 0 for failed requests (4xx/5xx)
- `session.totalCostCents` only incremented for successful requests
- Receipts are still inserted for all requests (audit trail preserved)

## Testing
Manual verification that:
- `costCents === 0` for 4xx/5xx responses
- `costCents > 0` for 2xx responses
- Receipts are inserted for all requests

Recommended: Add unit tests for 4xx/5xx scenarios.

### DIFF
--- a/packages/node/src/metering/seller-session-tracker.ts
+++ b/packages/node/src/metering/seller-session-tracker.ts
@@ -209,10 +209,15 @@ export class SellerSessionTracker {
     }
 
     if (this._receiptGenerator) {
+      // Only charge for successful requests (2xx/3xx status codes)
+      // Failed requests (4xx/5xx) get zero-cost receipts for audit trail
+      const isSuccess = statusCode >= 200 && statusCode < 400;
       const estimatedCostUsd =
-        (tokens.inputTokens * providerPricingUsdPerMillion.inputUsdPerMillion +
-          tokens.outputTokens * providerPricingUsdPerMillion.outputUsdPerMillion) /
-        1_000_000;
+        isSuccess
+          ? (tokens.inputTokens * providerPricingUsdPerMillion.inputUsdPerMillion +
+              tokens.outputTokens * providerPricingUsdPerMillion.outputUsdPerMillion) /
+            1_000_000
+          : 0;
       const effectiveUsdPerThousandTokens =
         tokens.totalTokens > 0 ? (estimatedCostUsd / tokens.totalTokens) * 1000 : 0;
       const unitPriceCentsPerThousandTokens = Math.max(0, effectiveUsdPerThousandTokens * 100);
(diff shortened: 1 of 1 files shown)


kotevcode:
on other services you are not getting charged if a call failed?..

PR #632, opened 2026-06-03
feat(contracts): Add recognized usage rewards

kotevcode:
## Description

Adds the recognized-usage rewards stack for ANTS emissions, including seller pools keyed by agent id, weighted buyer/seller/operator accounting, and program-gated emission minting. Also removes the deprecated SubPool client/config/docs surface so subscription pools are no longer advertised or callable.

## Release Notes

- Added stake-backed recognized usage rewards for seller pools, buyer usage, and seller operator usage
- Added lANTS stake receipts, bootstrap commitments, reward restaking, APY caps, and wash-trading resistance tests
- Removed the deprecated subscription pool command, client, config, and documentation

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Local Verification

- `cd packages/contracts && forge test`
- `pnpm --filter @antseed/node test -- buyer-payment-negotiator.test.ts`
- `pnpm --filter @antseed/cli typecheck`
- `pnpm --filter @antseed/website typecheck`


kotevcode:
@claude

kotevcode:
@claude

kotevcode:
Draft AIP for the recognized-usage trust design has been merged here:

https://github.com/AntSeed/AIPs/pull/1

Please review it alongside this implementation PR for spec/implementation alignment.

Augustas11 (commented):
Reviewed AIP-2 against this implementation for spec/implementation alignment, as invited — two independent reviewers traced all 102 normative clauses, plus a targeted follow-up on the handful we disagreed on.

Headline: this reads as a **faithful realization of the spec**. We found no clause we could confirm as divergent across the set, and no P0/P1 correctness defect in the new contract code — supply conservation across the cutover, budget freezing, next-epoch share checkpointing, zero-denominator handling, settlement-path resilience, and historical-power preservation all check out. Two things worth surfacing: one implementation heads-up and one spec-text clarification. Neither is a blocker.

---

**1. `AntseedEmissionsGate` — removing a minter strands its already-earned emissions (and the NatSpec says otherwise).**
For a non-destructive change, `setMinterController` rotates a bucket's controller without touching its budget or checkpoints — but `_removeMinter` does `delete _minters[id]` / `delete controllerMinterIds[controller]` in the same call in which it records the (correct) next-epoch share drop. Because claims are finalized-epoch-only and `_chargeMinterBudget` authorizes against the *live* `_minters[id].controller`, after removal the former controller — and any controller-mediated batch/remainder path — reverts `NotEmissionMinter` when trying to claim budget it already earned in prior finalized epochs. The preserved share checkpoint doesn't help, since the charge path gates on the deleted struct. It's recoverable (re-adding the same id restores claimability, so no permanent loss), but the NatSpec ("the in-progress epoch keeps its share so already-earned claims stay mintable") is false as written. Worth either blocking removal while unclaimed finalized budget exists, authorizing removed controllers against the historical checkpoint, or at minimum correcting the comment.

**2. The protocol-immutable vs governance-tunable boundary isn't written down (a documentation ask, not a change request).**
The AIP fixes the emission schedule and caps on-chain but doesn't say which of `minimumAccountedPoolPower`, `stakeActivationDelay`, the slash bps, and the dynamic-share configs are immutable vs owner/governance-tunable — the impl makes them all `onlyOwner`. A short per-parameter table — authority (immutable / owner / governance), permitted range, and whether a change affects only future epochs/positions or existing state — would let implementers and integrators reason about the trust surface without reading the deploy script. (For instance, `maxSlashBps` / `minEarlyExitSlashBps` are evaluated when an existing position withdraws, so their mutability has retroactive reach worth stating explicitly.)

---

Verdict: comment, not a change request — the implementation tracks the spec closely, and item 2 is a clarification rather than a defect. Happy to re-review any follow-up.

PR #650, opened 2026-06-10
feat(payments): Add verified usage manifests

kotevcode:
## Description

Adds buyer-verified off-chain usage manifests for payment stats. Sellers publish cumulative usage pointers in `NeedAuth`, buyers independently reconstruct and verify them before signing v2 metadata, and network-stats materializes verified manifests into service-level totals.

The buyer now persists verified usage records locally so cumulative pointer verification survives restarts. If pointer verification fails, payment falls back to legacy metadata signing instead of stalling.

## Release Notes

- Added buyer-verified usage manifest pointers for payment stats
- Added durable buyer and seller usage manifest files per channel
- Added network-stats materialization of verified per-service usage totals
- Preserved legacy metadata signing as a fallback when pointer verification fails

## Types of Changes

## Checklist

### DIFF
--- a/apps/network-stats/src/index.ts
+++ b/apps/network-stats/src/index.ts
@@ -7,6 +7,7 @@ import { NetworkPoller } from './poller.js';
 import { createServer } from './server.js';
 import { SqliteStore } from './store.js';
 import { MetadataIndexer } from './indexer.js';
+import { UsageManifestFetcher } from './usage-manifest-fetcher.js';
 
 const PORT = parseInt(process.env['PORT'] ?? '4000', 10);
 const CACHE_PATH = process.env['CACHE_PATH'];
@@ -39,6 +40,7 @@ if (chainConfig.statsContractAddress && typeof chainConfig.statsDeployBlock === 
   indexer = new MetadataIndexer({
     store,
     statsClient,
+    usageManifestFetcher: new UsageManifestFetcher(),
     chainId: CHAIN_ID,
     contractAddress: chainConfig.statsContractAddress.toLowerCase(),
     deployBlock: chainConfig.statsDeployBlock,
--- a/apps/network-stats/src/indexer.test.ts
+++ b/apps/network-stats/src/indexer.test.ts
@@ -1,9 +1,10 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
 
-import type { DecodedMetadataRecorded } from '@antseed/node';
+import type { DecodedMetadataPointerRecorded, DecodedMetadataRecorded, UsageManifest } from '@antseed/node';
 import { SqliteStore } from './store.js';
 import { MetadataIndexer } from './indexer.js';
+import type { UsageManifestFetcher } from './usage-manifest-fetcher.js';
 
 type MockStatsClient = Pick<
   import('@antseed/node').StatsClient,
@@ -15,6 +16,7 @@ type MockStatsClient = Pick<
 function makeMockClient(opts: {
   blockNumber: number;
   events?: DecodedMetadataRecorded[];
+  pointerEvents?: DecodedMetadataPointerRecorded[];
   throwOnFetch?: boolean;
 }): {
   client: MockStatsClient;
@@ -31,7 +33,7 @@ function makeMockClient(opts: {
       return opts.events ?? [];
     },
     async getMetadataPointerRecordedEvents() {
-      return [];
+      return opts.pointerEvents ?? [];
     },
   };
   return { client, fetchCalls };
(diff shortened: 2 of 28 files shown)


kotevcode:
## Design proposal: buyer-authored usage leaves with a chained root

*(Revised — supersedes my earlier comment, which folded in a gap around sparse signatures.)*

The architecture in this PR has both sides independently reconstructing a supposedly byte-identical manifest, with the buyer signing only if the roots happen to match. That makes verification fragile: any small disagreement permanently diverges the channel into legacy fallback with no resync path. Concretely:

- The 2s observation-wait timeout, a restart mid-request, or a service-name formatting difference fails verification once — and from then on the buyer's record list lags the seller's forever.
- An overdraft-capped signing (`effectiveAmount != requiredCumulativeAmount`) skips the commit while the seller has already appended the record.
- **Sparse signatures**: NeedAuth doesn't produce a fresh signature on every request. If the buyer has pre-signed ahead (per-request estimated auths with overdraft), the post-response NeedAuth arrives stale (`required <= currentCumulative`), `handleNeedAuth` returns early, and the verified record is never committed — but the seller appended it. Next pointer: root mismatch, permanent divergence, under completely normal operation.

It also forces both sides to hold full history (the buyer retains full request/response bodies in `BuyerUsageVerifier._observations` until a pointer arrives — which never happens against legacy sellers), the full-manifest rehash on every request is O(n²) over the channel lifetime, and the raw-leaf CID breaks once the manifest exceeds the IPFS 256 KiB chunk threshold (the gateway fetch then 404s forever).

The underlying principle worth changing: **the signer should author what it signs.** The buyer's signature is what makes the manifest trustworthy, so the buyer should be the source of the bytes instead of racing to reconstruct the seller's.

### Proposed shape

**1. Buyer authors the leaf.** After each response completes, the buyer builds the usage record from data it definitionally has — its own request body, the response it received, its own usage parse, and the cost it agreed to pay:

```jsonc
// UsageLeaf (same fields as UsageManifestRecord)
{
  "requestId": "...",
  "service": "gpt-5.5",
  "costUsdc": "1050",              // buyer-asserted (it's the payer)
  "cumulativeCostUsdc": "...",
  "inputTokens": "...", "cachedInputTokens": "...",
  "freshInputTokens": "...", "outputTokens": "...",
  "inputSha256": "...", "outputSha256": "..."
}
```

This eliminates the observation map, the 2s wait, and the body retention entirely — and the buyer keys any local persistence by its own session id, which also closes the unvalidated-`payload.channelId` filesystem path issue on the buyer side.

**2. Chained root instead of full-manifest hash.**

```
root_0 = 0x00…00
root_n = keccak256(root_{n-1} || sha256(canonicalJson(leaf_n)))
```

Verifying a new root needs only the previous *signed* root plus the new leaves — no full history on either side, O(1) per leaf. Publication becomes per-settlement leaf batches instead of one ever-growing file, so each published object stays under the raw-leaf size limit and the locally computed CID always matches what the IPFS node pins.

**3. Leaves ride the existing roundtrip, batched.** Signatures are sparser than requests (pre-signed estimates, overdraft), so the unit is *leaves since the last signed root*, not one leaf. `NeedAuth` drops the pointer fields; the SpendingAuth payload gains:

```jsonc
{
  "usageLeaves": [ /* all completed requests not yet covered by a signed root */ ],
  "usageRoot": "0x…"   // chain extended over every leaf, in order
}
```

with v2 metadata = `abi.encode(2, prevRoot, usageRoot)` (or a CID-of-batch form). Whenever the buyer signs — pre-request estimate, NeedAuth response, top-up, close — it attaches everything pending. In steady sequential use that's exactly one leaf per auth (the auth for request n+1 carries the leaf for request n); after pre-signed stretches it's a few. Every request gets committed to the chain regardless of how sparse the signatures are. Zero extra round trips, ~300 bytes per leaf.

Note the chain intentionally lags the in-flight request: a pre-request auth can't include the leaf for the request it's paying for (no response yet). Each signed root covers all *completed* requests; the in-flight one lands in the next auth.

**4. Seller verifies before accepting.** The verification burden flips to the seller, which is the right place because the seller submits on-chain and has the last word. On receiving a SpendingAuth the seller checks, per leaf:

- `inputSha256` / `outputSha256` match the bodies it served
- token counts match its own `parseResponseUsage` (exact, or within the existing `costTolerance`)
- `costUsdc` ≥ what it metered (or within tolerance)
- `usageRoot` chains correctly from the last root it accepted a signature for
- last leaf's `cumulativeCostUsdc` ≤ the signed `cumulativeAmount` (signed amount may run ahead due to estimates/overdraft; equality is only enforced on the final pre-settle auth — see 6)

If all pass, it persists the leaves and uses that auth for v2 settle. If not, it treats the auth as v1-only (or refuses further service). A buyer can never defame a seller's stats — only decline to contribute to them — and the worst case equals today's fallback: payments still flow, stats are just missing.

**5. Divergence is impossible by construction.** Only buyer-signed leaves enter the manifest, so there is no "seller appended a record the buyer never signed" state. After any hiccup, the next signed `(prevRoot, leaves, newRoot)` is self-contained — the chain continues from the last mutually signed root.

**6. Seller uploads at settle time, after a final auth.** Because the chain lags in-flight requests, the latest signed root may be a leaf or two behind what the seller is charging for. Mid-stream that's fine — only the auth submitted at settle/close matters for stats, intermediate roots are superseded. Before settling, the seller elicits one final auth via the existing post-response NeedAuth / close flow, whose root covers everything charged. If the buyer disappears first, the seller settles with the last signed root: stats undercount the tail requests, but payment is fully covered by `cumulativeAmount` — the right asymmetry (payment exact, stats best-effort). The seller (who benefits from the stats and runs persistent infra) publishes the leaf batch to IPFS; the contract change in this PR (emit `MetadataPointerRecorded`, defer verification off-chain) carries over unchanged.

**7. Indexer changes.** Instead of fetching one cumulative manifest and diffing channel totals, the indexer fetches the leaf batch for each pointer, recomputes the chain from the channel's last processed root, and applies per-leaf deltas. Monotonicity is enforced by the chain itself (each pointer must extend the previous root) — stronger and simpler than the current "new totals ≥ old totals" heuristic. The pending/retry queue from this PR stays, but should gain an attempt counter so a permanently bad pointer is dead-lettered rather than retried every tick and head-of-line blocking the `LIMIT 100` window.

### Trade-offs

- Sellers gain a verification obligation — but it's the same comparison logic the buyer performs in this PR, relocated to the side that already has both bodies and its own metering in hand at that moment.
- SpendingAuth payload grows by ~300 bytes per pending leaf (typically one).
- A hostile buyer can withhold or understate leaves to deny a seller stats — but that's equally true today (the buyer can always decline to sign v2), and payment correctness is unaffected since it's protected by `cumulativeAmount` independently.

Happy to discuss — the contract and indexer scaffolding in this PR are reusable as-is; the change is mostly in who builds the record, how the root is computed, and where the leaves travel.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #660, opened 2026-06-13
fix(desktop): harden bash auto-allow against interpreter command escapes

Augustas11:
Three intra-command escapes from the `bash:read` auto-allow surface that bypass `hasUnsafeShellComposition` / `hasShellWriteOperator` by living inside an allowlisted command's own argument grammar — same threat class as the composition-level holes closed in #573 / `128dab9a`, one layer deeper.

## What

`isReadOnlySegment` in [`apps/desktop/src/main/chat-tool-policy.ts`](https://github.com/AntSeed/antseed/blob/main/apps/desktop/src/main/chat-tool-policy.ts) allowed three commands whose argument grammars contain shell-exec or arbitrary-file-write primitives:

- **awk** — `system("…")`, `"cmd"|getline`, `{print | "sh"}`
- **sed** — `s///e` flag, `1e cmd` / `1,$e cmd` commands, `-f script.sed` (script from unverifiable file)
- **find** — `-fprintf path content`, `-fprint path`, `-fprint0 path`, `-fls path` (writes arbitrary content to arbitrary path with no `>` operator)

All three classified as `bash:read` → `isToolPermissionAutoAllowed` → silent execution, no approval prompt, no renderer event.

## Repro (against unpatched main, classifier harness)

```
awk 'BEGIN{system("env")}'      → bash:read → AUTO
sed -n '1e id' f                → bash:read → AUTO
find . -fprintf /tmp/x '%p\n'   → bash:read → AUTO
```

## Fix

Extend the existing guard alternations in `isReadOnlySegment` (mirrors the maintainer's blocklist-extension pattern from `128dab9a`):

- `find`: add `-fprintf|-fprint|-fprint0|-fls`
- `sed`: add the `e` command/flag and `-f` script-from-file
- `awk`: add `system(`, `getline`, and `| "…"` / `"…" |` pipe-to-shell forms

Total: 4 lines changed in `chat-tool-policy.ts` (one comment block + the three regex extensions).

## Tests

Four new `node:test` suites in `chat-tool-policy.test.ts`:

- `awk system()/getline/shell-pipe must not auto-allow` — 3 PoCs
- `sed e command/flag and -f script-from-file must not auto-allow` — 5 PoCs
- `find -fprintf / -fprint / -fls must not auto-allow` — 4 PoCs
- `benign awk/sed/find invocations stay auto-allowed` — 7-case regression guard so ordinary `awk '{print $1}'`, `sed 's/foo/bar/' f`, `find . -name '*.ts'`, etc. still classify as `bash:read`

`npm test` in `apps/desktop`: **12/12 pass** locally.

## Lineage

Continuation of the chat-safety auto-allow arc: #592 (initial guard), #631 (zero-width-bypass fix), #642 (inline-blur hotfix), #643 (lowercase base58), #573 / `128dab9a` (composition-level intra-command fix). This PR closes the equivalent escapes at the *builtin-command* level rather than the composition level.

Internal triage IDs in this fork's audit: BUG-200 (awk, High), BUG-201 (sed, High), BUG-202 (find, Medium).

### DIFF
--- a/apps/desktop/src/main/chat-tool-policy.test.ts
+++ b/apps/desktop/src/main/chat-tool-policy.test.ts
@@ -36,6 +36,87 @@ test('read-looking shell commands with mutation operators are not auto-allowed',
   assert.equal(describeBash('sed -i s/foo/bar/g file.txt').permissionKey, 'bash:other');
 });
 
+test('awk system()/getline/shell-pipe must not auto-allow', () => {
+  // POSIX awk's system() and getline-via-pipe execute arbitrary shell. The
+  // first-word allowlist would otherwise classify these as `bash:read` and
+  // skip the approval prompt entirely.
+  const bypasses = [
+    "awk 'BEGIN{system(\"env\")}'",
+    "awk 'BEGIN{\"id\" | getline x; print x}' /dev/null",
+    "awk '{print | \"sh\"}' file",
+  ];
+  for (const command of bypasses) {
+    const approval = describeBash(command);
+    assert.equal(
+      isToolPermissionAutoAllowed(approval.permissionKey),
+      false,
+      `expected ${command} to require explicit approval, got ${approval.permissionKey}`,
+    );
+  }
+});
+
+test('sed e command/flag and -f script-from-file must not auto-allow', () => {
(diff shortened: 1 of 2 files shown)


kotevcode:
@claude

Issue #504, opened 2026-05-12
Proposal: Model Verification via Behavioral Fingerprinting


Augustas11:
## Problem

A provider can advertise `claude-sonnet-4-6` on the DHT but silently serve a cheaper model — a nano-tier endpoint, a local Ollama quantization, or a completely different model family. The buyer pays the premium price and receives degraded output.

The current reputation system doesn't catch this. Ghost channel counts, ERC-8004 feedback, and volume all track *whether* delivery happened — not *what* was delivered. A cheating provider scores just as well as an honest one.

I run a provider node on the network and this was surfaced as a real user concern.

## Proposed Solution

Different LLM families and tiers produce measurably distinct responses to structured prompts at `temperature=0` — format compliance precision, arithmetic accuracy, self-reported identity, language adherence. These differences are consistent and hard to spoof without actually running the target model.

**Behavioral fingerprinting:** the router periodically injects small probe requests alongside real traffic. Probes have automated scorers (pure functions, no human judgment). A session runs N probes → produces an aggregate score → feeds into peer selection and optionally ERC-8004 `Accuracy` signal.

```
onResult() hook
    └── every N requests: fire probe session async (never blocks real traffic)
          └── selectRandomProbes(advertisedModel)
                └── probe.matcher(response, advertisedModel) → 0–1 score
                      └── aggregateScore < threshold → fail verdict
                            └── filter peer from selectPeer() + optional ERC-8004 Accuracy: 0
```

## Why Not ZK Weight Proofs?

ZK proofs of model weights (e.g. Dolphin's approach) are the gold standard but require proving hardware and add 100–500ms overhead per request. Too costly for general inference at this stage.

Behavioral fingerprinting:
- Zero provider-side changes required — honest providers pass automatically
- Zero overhead on real requests — probes are separate cheap side-channel calls (≤30 output tokens each)
- Catches the common case: casual model substitution for profit

The two approaches are complementary — fingerprinting now, optional TEE/ZK attestation tier later for high-trust routes.

## Sketch of What I'd Implement

**`packages/node/src/verification/`**
- `types.ts` — `ModelProbe`, `VerificationSession`, `VerificationConfig`
- `fingerprint-db.ts` — probe database with matchers (format, math, identity, language)
- `verifier.ts` — `ModelVerifier` class: session management, rolling per-peer scores

**`packages/router-core/src/verification-middleware.ts`**
[...]

PR #661, opened 2026-06-13
docs(protocol): add model-verification design spec

kotevcode:
## What

Adds `docs/protocol/spec/07-model-verification.md` — a design spec for verifying that the model a Seller serves matches the model it advertised on the DHT, and for escalating a confirmed substitution to on-chain slashing. Wires it into the spec README (layer 7, doc table, version 1.2).

Status is **proposed / design** — nothing here is implemented. The goal is to give contributions (notably #504) a common architecture to target.

## Why

The reputation and metering layers answer *whether* a request was served and *how much* was billed — never *what* was served. A Seller can advertise `claude-sonnet-4-6` and silently serve a nano-tier endpoint or a quantized copy; settlement volume, ghost-channel rate, and ERC-8004 feedback score the cheat exactly like an honest peer.

## Approach

Layered `unattested → signed → attested` ladder:

| Mechanism | Role |
|---|---|
| M1 Behavioral probes | cheap deterrent / triage only — **not** penalty-driving |
| M2 Reference shadow sampling | core: duplicate ~2% of real traffic to a reference, compare distributionally over N≥30 samples |
| M3 Signed responses | non-repudiable evidence so verdicts are portable to a dispute |
| M4 Passive fingerprinting | free triage to aim the sampling budget |
| M5 TEE attestation | premium tier; doubles as the M2 reference |

Recommended v1 is **M2 + M3 + M4**.

## Relationship to #504

The behavioral-fingerprinting proposal in #504 is captured as **M1**. The spec credits its strengths (zero Seller-side change, honest peers pass automatically, catches casual substitution) but documents why it can't stand alone:

- **Distinguishability** — a static probe bank sent over a side channel is a defeat device once it's public; the spec's central design principle is that any signal distinguishable from real traffic gets gamed.
- **Tier ≠ identity** — capability probes separate nano from frontier but can't catch a similar-capability impersonator or quantization.
- **Self-reported identity** is dropped as a signal (one-line system prompt defeats it; training contamination makes it noisy).

This reframes #504's `Accuracy: 0`-on-fail idea: on-chain consequences come only after arbiter confirmation (re-query against a reference), never directly from a probe verdict. The middleware skeleton and `sendDirectToPeer()` integration question from #504 carry over largely unchanged.

Refs #504

### DIFF
--- a/docs/protocol/spec/07-model-verification.md
+++ b/docs/protocol/spec/07-model-verification.md
@@ -1,0 +1,278 @@
+# 07 - Model Verification
+
+**Status:** Proposed / design. None of the mechanisms below are implemented yet.
+This document specifies the design space and the recommended architecture so that
+implementations (and contributions such as issue #504) target a common model.
+
+## Overview
+
+The reputation layer ([05-reputation.md](./05-reputation.md)) and metering layer
+([03-metering.md](./03-metering.md)) answer **whether** a request was served and
+**how much** was billed. Neither answers **what** was served.
+
+A Seller can advertise a premium model (e.g. `claude-sonnet-4-6`) on the DHT and
+silently serve a cheaper substitute — a different model family, a nano-tier
+endpoint, or an aggressively quantized copy of the correct model. The Buyer pays
+the advertised price and receives degraded output. Settlement volume,
+ghost-channel rate, and ERC-8004 feedback all score the cheating Seller exactly
+as well as an honest one, because delivery still happened.
+
+**Model verification** is the set of mechanisms that let a Buyer gain confidence
+that the model it paid for is the model it received, and — where the evidence is
+strong enough — escalate a confirmed substitution to on-chain slashing.
+
+There is no central verification authority. Every mechanism here is either
+Buyer-local or settled through existing peer-to-peer and on-chain primitives.
+
+---
+
+## Threat Model
+
+The adversary is a profit-motivated Seller that:
+
+- Sees every request in plaintext before responding.
+- Can run more than one backend and route each request to whichever backend it
+  chooses, based on the request content.
+- Will serve the advertised model only when it believes it is being watched, and
+  the cheap substitute otherwise.
+
+This last capability is the crux. **Any verification signal that is
+distinguishable from ordinary Buyer traffic is a defeat device waiting to
+happen**: the Seller detects the probe, serves the real model for that request,
+and cheats everything else. Distinguishability is the property to design out, not
+a detail to tune later.
+
+A secondary adversary is a malicious **Buyer** that fabricates or cherry-picks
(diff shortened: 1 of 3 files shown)


Augustas11 (commented):
As the author of issue #504 (which the spec explicitly cites as a contribution target), posting two items I'd want pinned down before targeting an implementation against it — both surfaced by a local two-reviewer pass on head `c9b936e`. Direction reads sound and framing distinguishability as the load-bearing principle is the right anchor.

## 1. Does the threat model cover coordinated multi-buyer adversaries?
**Spec text** (line 232 of `07-model-verification.md`): "This is countered by requiring the exhibit's sample set to reconcile against the Buyer's metered request history ... and by the arbiter's own fresh re-query in step 9, which does not depend on the Buyer's samples at all."
**As #504 author, what I'd want pinned down**: The malicious-Buyer treatment is single-actor. Two open cases an implementer would have to invent a defense for: (i) N sybil buyers each running real, paid sessions against an honest Seller and pooling their M2 samples to manufacture a distributional fail — every sample IS genuinely metered, so "reconcile against metered history" passes; (ii) the symmetric case where a substituting Seller colludes with sybil buyers to manufacture exonerating samples that average out a real substitution. Step-9 re-query catches (i) only while substitution is live; (ii) survives entirely.
**Suggested clarification**: Could the threat model say whether coordinated multi-buyer evidence-pooling (accusatory or exculpatory) is in scope for v1, and whether accusation weighting is per-buyer, stake-weighted, or left to the arbiter?

## 2. What's the full preimage for `responseAuth`?
**Spec text** (line 125): "`responseAuth = sign_secp256k1( DOMAIN_TAG || keccak256(requestBytes) || keccak256(responseBytes) )`"
**As #504 author, what I'd want pinned down**: The signed tuple binds only `(who, request-bytes, response-bytes)`. It does not bind the advertised model id, channel/session id, buyer identity, or a nonce/timestamp — so two buyers issuing a byte-identical request (short prompts, the M2 reference shadow itself, common benchmarks) receive interchangeable signatures, and a Seller advertising two models can argue an exhibit applies to the cheaper one. `DOMAIN_TAG` itself is unspecified relative to existing EIP-712 type-hashes in `AntseedChannels.sol`, which is a cross-protocol replay surface on the same key. An implementer of #504 has to invent the canonical claim envelope.
**Suggested clarification**: Could the spec specify the preimage fields (advertised-model-id, channel/session id, buyer, nonce/timestamp) and `DOMAIN_TAG` encoding — or state that these bindings are inherited from the metering receipt and reference the carrying fields?

Happy to be wrong on either of these — would help me scope an issue-#504 implementation PR if/when you signal it's time. No verdict gate — COMMENT review.

kotevcode:
hehe yes don’t worry. The agents are here to set the overall guidance and direction. I’m here to make sure it will actually work :)

PR #688, opened 2026-06-16
Buyer service plugins + gasless USDC auto-deposit (Circle Paymaster + EIP-7702)

gabrielpoca:
## What this does

Introduces a third plugin type, **`service`**, alongside `provider` and `router`, and ships the gasless **auto-deposit** funder as the first service plugin. A service is a host-driven background task that runs on the buyer's behalf and declares the sensitive capabilities it needs (e.g. `wallet`, `chain`), which the host grants explicitly. The desktop gets a generic **Services** settings panel to view and toggle them.

This started as a single auto-deposit feature and grew into a reusable plugin contract so future buyer-side services (not just funding) can plug in the same way.

### New `service` plugin type (`@antseed/node`)
- `AntseedServicePlugin` added to the plugin union, with a `kind`, declared `capabilities`, and a `createService(host)` factory.
- Generic capability injection: the plugin declares what it needs; the host grants only those (`wallet` -> identity key, `chain` -> chain config). A service never receives a capability it did not request, and is skipped (with a reason) when a required capability is unavailable.
- New shared package **`@antseed/service-core`** with the runtime contracts and `requireWallet`/`requireChain` helpers.

### Auto-deposit as an installable plugin (`@antseed/service-auto-deposit`)
- Moved from `packages/auto-deposit` into `plugins/service-auto-deposit`; it is now loaded dynamically from the plugins dir like provider/router plugins (no longer compiled into the CLI).
- Sweeps loose USDC in the buyer wallet into `AntseedDeposits` with **no ETH**: gas is paid in USDC via the Circle Paymaster, transactions are relayed by the Pimlico bundler, and the wallet is upgraded once via EIP-7702 on the first deposit.
- Respects the on-chain credit limit (deposits beyond it are left as loose funds, never lost) and a circuit breaker that pauses on deterministic failures.

### CLI / buyer proxy
- Dynamic service-plugin loading (`registry`, `loader`, `services`); services load only when settlement is enabled.
- `GET/POST /_antseed/services` control endpoint (loopback-guarded): the GET merges the trusted catalog with live plugins so a service that is available but not yet running is still listed with a reason; the POST persists consent and activates/pokes the service.

### Desktop
- New generic **Services** panel (replaces the funding-only block): each service shows a status badge (Active / Inactive / Not installed), a toggle, its live status, and for funding the receive address + QR.
- The deposit warning is driven by the live on-chain credit-limit headroom rather than a hardcoded cap.
- Friendly "Auto Deposit" label is shown in every state, including before the plugin is installed.
- The service is seeded on launch and the buyer spawn waits on plugin readiness so a freshly installed service is not missed until a restart. Production bundling mirrors the router (offline install).

## Testing

Exercised end to end on Base mainnet: the buyer proxy loads the service, the Services endpoint reports the live deposit limit, real USDC sent to the wallet was swept into `AntseedDeposits` gaslessly, and the funds were later withdrawn gaslessly back out. Unit tests cover the deposit planner, status mapping, and the credit-limit headroom.

## Follow-ups

- `@antseed/service-core` and `@antseed/service-auto-deposit` are not published to npm yet. The desktop installs trusted plugins from the registry (dev) or the app bundle (production), so both packages need publishing for the dev install path to work without the app bundle.
- `CHANGELOG.md` is intentionally left untouched until those packages are published.
- The 10 USDC figure in earlier discussion is the base credit limit; the actual cap grows per buyer, which is why the UI reads it live.

### DIFF
--- a/apps/cli/src/proxy/buyer-proxy.ts
+++ b/apps/cli/src/proxy/buyer-proxy.ts
@@ -6,6 +6,8 @@ import { join } from 'node:path'
 import {
   computeOnChainReputationScore,
   type AntseedNode,
+  type AntseedServicePlugin,
+  type ChainConfig,
   type PeerInfo,
   type PeerMetadata,
   type RequestStreamResponseMetadata,
@@ -13,6 +15,10 @@ import {
   type SerializedHttpRequest,
   type SerializedHttpResponse,
   type SerializedHttpResponseChunk,
+  type Service,
+  type ServiceCapabilities,
+  type ServiceCapability,
+  type ServiceHost,
 } from '@antseed/node'
 import {
   createOpenAIChatToAnthropicStreamingAdapter,
@@ -51,6 +57,7 @@ import {
   attachStreamingAntseedHeaders,
 } from './telemetry.js'
 import { DEFAULT_BUYER_PEER_REFRESH_INTERVAL_MS } from '../config/defaults.js'
+import { loadConfig, saveConfig } from '../config/loader.js'
 
 // Re-export for backward compatibility (used by tests and other consumers)
 export { selectCandidatePeersForRouting, type CandidatePeerRouteSelection } from './routing.js'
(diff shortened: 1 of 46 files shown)


apps/cli/src/proxy/buyer-proxy.ts
```diff
@@ -550,6 +585,30 @@ export class BuyerProxy {
     this._node.startBackgroundPeerDiscoverySweep()
   }
 
+  private _startAutoDeposit(): void {
+    const chain = this._chainConfig
+    const identity = this._node.identity
+    if (!chain?.autoDeposit || !identity) return
+    const manager = createAutoDepositManager({
+      chain,
+      privateKey: identity.wallet.privateKey as `0x${string}`,
```
naps62: one more thing: following this code path, *I think* this ends up holding a raw pkey in memory indefinitely. no idea if that's something that already happens elsewhere, but are we ok with that from a security perspective?

I don't know enough about the app to judge this

gabrielpoca: hummm good point. Let me check. I think this app already has the private key in mem all the time because it's signing things constantly, but I'm not 100% certain.

gabrielpoca: So the key is already in memory in other places, but on top of that, it's also plaintext on disk. Because this app is auto-signing, we probably cannot avoid having it in memory all the time, but we could encrypt on disk I guess, like wallets do. I think this is not a blocker for this PR cc @kotevcode

PR #708, opened 2026-06-21
fix(desktop): Surface updater failures

### LINKED ISSUE #604: Surface desktop auto-updater errors in the UI
The desktop updater can currently fail in ways that look like “nothing happened” to the user, especially on macOS when clicking the update/restart button.

**Problem**
The updater pipeline logs errors in the main process, but the renderer does not surface failures to the user. The install button handler also ignores the returned Promise:

- `autoUpdater.on('error', ...)` logs to console only
- `ipcMain.handle('app:install-update', () => { autoUpdater.quitAndInstall(false, true); })` does not report success/failure/status
- `TitleBar.handleUpdate()` calls `bridge?.installUpdate?.()` and ignores rejection/status

So if `quitAndInstall()` throws, refuses, or gets blocked, the user just sees no visible feedback.

**Likely user-facing scenario**
A macOS user clicks the update button and nothing appears to happen. This may be caused by launching from the DMG instead of `/Applications`, quarantine/translocation, permissions, multiple app instances, signature/arch mismatch, or shutdown cleanup blocking the updater path — but the app gives no actionable error.

**Expected behavior**
When update download or install fails, the desktop UI should show a clear, non-scary message such as:

- “Update failed to install. Quit AntSeed, move it to Applications, reopen, and try again.”
- Include a “Details” / copyable error message for support.
- Keep enough diagnostic context in logs for debugging.

**Implementation notes**
- Send updater errors to renderer via `app:update-status` with an `error` state.
- Wrap/report `app:install-update` failures.
- Have `TitleBar` display an error state instead of silently ignoring `installUpdate()` rejection.
- Consider detecting macOS DMG/translocation/non-Applications launch path and showing a targeted hint.
- Verify the `before-quit` cleanup path does not interfere with `autoUpdater.quitAndInstall(false, true)`.

kotevcode:
## Description

Surfaces Desktop auto-updater failures in the renderer instead of only logging them in the main process. Also fixes the macOS Quit menu flow so the first quit action exits after runtime cleanup instead of requiring a second click.

Fixes #604.

## Release Notes

- Fixed Desktop auto-update failures so download and install errors appear in the title bar with copyable details.
- Fixed macOS Quit so the first menu action exits after child process cleanup instead of requiring a second click.

## Types of Changes

## Checklist

- [ ] I have added tests (if appropriate)

## Verification

- `pnpm -C apps/desktop run build:main`
- `pnpm -C apps/desktop run typecheck:renderer`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ### Fixed
 
+- Fixed Desktop auto-update failures so download and install errors appear in the title bar with copyable details, and fixed macOS Quit so the first menu action exits after cleanup instead of requiring a second click.
 - Fixed buyer response-auth timeout warnings for non-inference probes and sellers that do not advertise response-auth support.
 - Fixed buyer discovery so temporarily unreachable metadata endpoints are probed for recovery before the full exponential cooldown expires, allowing recovered peers to reappear in buyer peer lists sooner.
 - Fixed Desktop chats for peers that disappear from discovery so the header reports that the peer was not found and disables the composer instead of showing stale peer identifiers.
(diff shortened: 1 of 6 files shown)


kotevcode:
@claude

PR #711, opened 2026-06-23
perf(desktop): Lazy-load renderer views

kotevcode:
## Description

Changes the Desktop renderer so only the active view is mounted, with route chunks lazy-loaded through a shared registry. Adds preloading for likely next views, a lightweight loading fallback for first navigation, retained view-local state helpers, and renderer tests for the new routing/state behavior.

## Release Notes

- Reduced Desktop renderer memory and initial bundle pressure by loading only the active view.
- Added Desktop route preloading and a lightweight loading state so first chat navigation does not appear blank while the chat view chunk loads.

## Types of Changes

- [x] Chore (maintenance tasks, refactoring, or non-functional changes)

## Checklist

Verification:
- `pnpm -C apps/desktop run test:renderer`
- `pnpm -C apps/desktop run typecheck:renderer`
- `pnpm -C apps/desktop run build:renderer`
- `pnpm -C apps/desktop run test`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ### Changed
 
+- Changed Desktop renderer navigation to load only the active view, preload likely next views, and show a lightweight loading state while lazily loaded pages resolve.
 - Reduced the default buyer response-auth evidence sample rate from 20% to 0.5% to limit local `verification_samples` growth during high-request sessions.
 - Increased the default free-usage on-chain record flush interval from 10 seconds to 5 minutes to reduce background transaction frequency while preserving batch, disconnect, and shutdown flushes.
 
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -18,7 +18,8 @@
     "build": "npm run build:main && npm run build:renderer",
     "build:main": "tsc -p tsconfig.main.json",
     "build:renderer": "vite build",
-    "test": "npm run build:main && node --test dist/main",
+    "test": "pnpm run build:main && node --test dist/main/*.test.js && pnpm run test:renderer",
+    "test:renderer": "vitest run --dir src/renderer",
     "typecheck:renderer": "tsc -p tsconfig.renderer.json --noEmit",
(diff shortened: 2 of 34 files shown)


kotevcode:
@claude

kotevcode:
Addressed the review feedback in `7a61e07e`:

- Stabilized `useUiSelector` with refs so inline selectors do not recreate the snapshot getter each render.
- Added selector regression coverage.
- Preloaded the initial Discover route along with Chat after setup exits.
- Documented the `ViewHost active={true}` invariant and renderer-lifetime retained caches.
- Added JSDoc for retained mutable value cloning.

Verified:
- `pnpm -C apps/desktop run test:renderer`
- `pnpm -C apps/desktop run typecheck:renderer`
- `pnpm -C apps/desktop run build:renderer`
- `git diff --check`

kotevcode:
Follow-up for the perceived action delay:

- Chat deletion now removes the conversation from renderer state before awaiting `chatAiDeleteConversation`, so the sidebar updates immediately instead of waiting on IPC/storage/attachment cleanup.
- The main-process delete and conversation-list refresh still run afterward to persist and reconcile state.
- Added a regression test that keeps the delete IPC promise pending and asserts the conversation is already removed from the UI list.

Verified:
- `pnpm -C apps/desktop run test:renderer`
- `pnpm -C apps/desktop run typecheck:renderer`
- `pnpm -C apps/desktop run build:renderer`
- `git diff --check`

kotevcode:
Additional latency audit:

- Found one more action with the same pattern: `abortChat` waited for `chatAiAbort` before clearing sending/disabled state.
- Updated abort to clear renderer sending state immediately, then await IPC.
- Added regression coverage for pending abort IPC while the UI is already cleared.
- Checked nearby actions: rename is already optimistic; open/create need main-process data; config save already sets saving/form state first; channel close only opens the payment portal.

Verified:
- `pnpm -C apps/desktop run test:renderer`
- `pnpm -C apps/desktop run typecheck:renderer`
- `pnpm -C apps/desktop run build:renderer`
- `git diff --check`

PR #713, opened 2026-06-26
feat(node): pluggable verifier sdks

essamhassan:
## What

Introduces Verifier SDKs as antseed plugins allowing sellers to offer verifications of their claims through verification SDKs

## Testing

- Added tests
- Ran test suite
    - pnpm --filter=@antseed/node typecheck
    - pnpm --filter=@antseed/cli typecheck
    - pnpm --filter=@antseed/cli test
    - pnpm --filter=@antseed/node test
- e2e test with first verifier SDK on TDX TEE hardware, running against basic dcap attestation

## Checklist

- [x] `pnpm run build` passes
- [x] `pnpm run test` passes

### DIFF
--- a/apps/cli/src/cli/commands/buyer/start.ts
+++ b/apps/cli/src/cli/commands/buyer/start.ts
@@ -11,11 +11,11 @@ import { AntseedNode, DepositsClient, getInstance, resolveChainConfig } from '@a
 import type { NodePaymentsConfig } from '@antseed/node'
 import { OFFICIAL_BOOTSTRAP_NODES, parseBootstrapList, toBootstrapConfig } from '@antseed/node/discovery'
 import { setupShutdownHandler } from '../../shutdown.js'
-import { loadRouterPlugin, buildPluginConfig, getPackageVersions } from '../../../plugins/loader.js'
+import { loadRouterPlugin, loadVerifierPlugin, buildPluginConfig, getPackageVersions } from '../../../plugins/loader.js'
 import { ensurePluginsUpToDate } from '../../../plugins/drift.js'
 import { resolvePluginPackage } from '../../../plugins/registry.js'
 import { BuyerProxy } from '../../../proxy/buyer-proxy.js'
-import { normalizeVerifierIds, type VerifierPolicy } from '../../../plugins/verifier.js'
+import { curatedVerifierIds, resolveVerifierPolicy, type VerifierPolicy } from '../../../plugins/verifier.js'
 import { resolveEffectiveBuyerConfig, type BuyerRuntimeOverrides } from '../../../config/effective.js'
 import type { BuyerCLIConfig } from '../../../config/types.js'
 
@@ -414,17 +414,31 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
       const proxyPort = effectiveBuyerConfig.proxyPort
       const proxySpinner = ora(`Starting local proxy on port ${proxyPort}...`).start()
       let verifierPolicy: VerifierPolicy | undefined
-      if (options.verifier === false) {
-        verifierPolicy = undefined
-      } else {
-        try {
-          verifierPolicy = {
-            prefer: normalizeVerifierIds(String(options.verifiers ?? '')),
-            require: Boolean(options.requireVerifier),
+      try {
+        verifierPolicy = resolveVerifierPolicy({
+          verifier: options.verifier,
+          verifiers: options.verifiers,
+          requireVerifier: options.requireVerifier,
+        })
+      } catch (err) {
+        console.error(chalk.red((err as Error).message))
+        process.exit(1)
+      }
+
+      // Keep the request path import-only; the seller-specific default is known per peer.
+      if (verifierPolicy) {
+        const toPreload = verifierPolicy.prefer?.length ? verifierPolicy.prefer : [...curatedVerifierIds()]
+        for (const id of toPreload) {
+          try {
+            await loadVerifierPlugin(id)
+          } catch (err) {
+            const msg = `Verifier "${id}" could not be prepared: ${(err as Error).message}`
+            if (verifierPolicy.require) {
+              proxySpinner.fail(chalk.red(msg))
+              process.exit(1)
+            }
+            console.warn(chalk.yellow(`${msg} — optional verification for this SDK will be skipped.`))
           }
-        } catch (err) {
-          console.error(chalk.red((err as Error).message))
-          process.exit(1)
         }
       }
 
--- a/apps/cli/src/cli/commands/buyer/start.ts
+++ b/apps/cli/src/cli/commands/buyer/start.ts
@@ -15,6 +15,7 @@ import { loadRouterPlugin, buildPluginConfig, getPackageVersions } from '../../.
 import { ensurePluginsUpToDate } from '../../../plugins/drift.js'
 import { resolvePluginPackage } from '../../../plugins/registry.js'
 import { BuyerProxy } from '../../../proxy/buyer-proxy.js'
+import { normalizeVerifierIds, type VerifierPolicy } from '../../../plugins/verifier.js'
 import { resolveEffectiveBuyerConfig, type BuyerRuntimeOverrides } from '../../../config/effective.js'
 import type { BuyerCLIConfig } from '../../../config/types.js'
 
@@ -198,6 +199,9 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
     .option('--disable-metadata-v2-services', 'runtime-only opt-out from per-service buyer metadata v2 attribution')
     .option('--peer <peerId>', 'pin all requests to a specific peer ID (40-char hex EVM address), bypassing the router')
     .option('--log-filter <sourceOrText>', 'show only debug logs matching a source or text, e.g. ProxyMux')
+    .option('--verifiers <ids>', "ordered, comma-separated verifier SDK ids to verify sellers with (default: the seller's advertised default if trusted)")
+    .option('--require-verifier', 'refuse to route unless a verifier SDK verifies the seller (default: verify but route anyway)')
+    .option('--no-verifier', 'disable seller verification entirely')
     .action(async (options) => {
       const globalOpts = getGlobalOptions(buyerCmd)
       const config = await loadConfig(globalOpts.config)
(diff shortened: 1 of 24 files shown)


apps/cli/src/cli/commands/buyer/start.ts
```diff
@@ -197,6 +198,9 @@ export function registerBuyerStartCommand(buyerCmd: Command): void {
     .option('--metadata-fetch-timeout-ms <number>', 'runtime-only timeout for each peer metadata HTTP fetch during discovery', Number)
     .option('--disable-metadata-v2-services', 'runtime-only opt-out from per-service buyer metadata v2 attribution')
     .option('--peer <peerId>', 'pin all requests to a specific peer ID (40-char hex EVM address), bypassing the router')
+    .option('--verifiers <ids>', "ordered, comma-separated verifier SDK ids to verify sellers with (default: the seller's advertised default if trusted)")
+    .option('--require-verifier', 'refuse to route unless a verifier SDK verifies the seller (default: verify but route anyway)')
+    .option('--no-verifier', 'disable seller verification entirely')
```
essamhassan, line 204: these defaults we need to agree on

packages/node/src/discovery/announcer.ts
```diff
@@ -75,6 +75,8 @@ export interface AnnouncerConfig {
   stakingClient?: StakingClient;
   reannounceIntervalMs: number;
   signalingPort: number;
+  /** Extra peer capability strings to advertise (merged with the built-in set). */
+  capabilities?: string[];
```
essamhassan, line 82: added these because did not fit in any existing field in announcement

packages/node/src/interfaces/plugin.ts
```diff
@@ -33,4 +33,53 @@ export interface AntseedRouterPlugin extends AntseedPluginBase {
...
+
+export interface VerifyContext {
+  peerId: string
+  /** Verifier id selected by the buyer; must match the SDK name. */
+  verifierId: string
+  /** Seller prover path for this verifier. */
+  attestPath: string
+  fetchFromSeller(req: SellerRequest): Promise<SellerResponse>
+  signal?: AbortSignal
+}
```
essamhassan, line 71: this is to allow for non TEE verifications to plug in

apps/cli/src/cli/commands/buyer/start.ts
kotevcode: how can the buyer know what verifiers exists? maybe we need to have a list of known sdks, and by default it will use them while the buyer can disable any of them and add any new one.

apps/cli/src/cli/commands/network/browse.ts
```diff
@@ -483,6 +486,14 @@ function renderCompactTable(peers: PeerInfo[], hasChainData: boolean): void {
       peer.displayName ?? chalk.dim('—'),
     );
     if (anyVerificationLinks) row.push(formatVerificationLinks(peer));
```
kotevcode, line 488: all of this must affect reputation, at the end the user should see higher reputation of sellers that can get verified

apps/cli/src/cli/commands/seller/start.ts
```diff
@@ -573,11 +575,37 @@ export function registerSellerStartCommand(sellerCmd: Command): void {
         ? { sellerContract: sellerContractCfg.address }
         : undefined
 
+      // Load provers before advertising verifier capabilities.
+      let requestedVerifierIds: string[]
```
kotevcode, line 582: we need to also add this to `seller setup` and use the same predefined list to choose from unless the sellers wants to add a custom proverPlugin

apps/cli/src/plugins/registry.ts
```diff
@@ -48,6 +50,13 @@ export const TRUSTED_PLUGINS: TrustedPlugin[] = [
     description: 'Local router for Claude Code, Codex',
     package: '@antseed/router-local',
   },
+  {
```
kotevcode, line 59: might be a good point in time to divide to separate files or vars `TRUSTED_PROVIDER_PLUGINS`, `TRUSTED_ROUTER_PLUGINS`, `TRUSTED_VERIFIER_PLUGINS`

apps/cli/src/proxy/buyer-proxy.ts
kotevcode, line 1: **Verifier plugins get an unrestricted request primitive that can spend buyer funds** (`buyer-proxy.ts:1210`)

The `reach` closure forwards the plugin's arbitrary `method`/`path`/`headers`/`body` into `node.sendRequest` — the same path used for paid dispatch. `attestPath` is passed as advice only; nothing pins the request to the attest route. That path has payment negotiation live: on a 402 it auto-runs `handle402`, opening a channel (on-chain ReserveAuth locking USDC) and signing SpendingAuths. A buggy or malicious verifier can call `reach({method:'POST', path:'/v1/chat/completions', ...})` and move real funds — even against a peer that then fails verification and is never used.

Fix at the harness layer: constrain `reach` to `ANTSEED_ATTEST_PATH/<chosenId>` and bypass payment negotiation for verifier traffic.

packages/node/src/seller-request-handler.ts
kotevcode, line 1: **Attest route is free, unauthenticated, unmetered, and un-rate-limited** (`seller-request-handler.ts:89`)

The `/_antseed/attest/<id>` route is dispatched before payment, session, free-usage, and load-accounting logic, and no lower layer (ProxyMux / connection-manager) imposes a per-peer request or concurrency cap. Any connected buyer can loop `POST /_antseed/attest/<id>`, each call driving an expensive TEE DCAP quote generation at zero cost — a cheap seller-CPU DoS that starves paying requests, invisible to the load-based announcer metadata.

Route attestation through the free-usage manager or add per-peer rate limiting at the handler layer.

apps/cli/src/proxy/buyer-proxy.ts
kotevcode, line 1: **Verification runs on every proxied request with no per-peer caching** (`buyer-proxy.ts:1220`)

The check sits in the per-request pinned-peer handler with only the immutable policy in scope — no outcome cache anywhere. Every request pays a full attestation round-trip (DCAP quote) before dispatch, and the first request to a given seller can block on a synchronous `npm install` (see the default-on install finding).

Cache the outcome per `(peerId, verifierId)` with a TTL tied to the existing peer-cache refresh, re-verifying only on expiry or metadata change.

apps/cli/src/plugins/verifier.ts
kotevcode, line 1: **Default-on verification triggers a blocking, timeout-less npm install from remote seller metadata** (`verifier.ts:90` → `loader.ts` → `manager.ts`)

`--no-verifier` makes verification default-on. With an empty `prefer` list, `selectVerifier` picks a curated seller default straight from DHT metadata, and `loadVerifierPlugin` → `ensureTrustedPluginInstallReady` → `installPlugin` runs `npm install` (no `timeout` option) inline in the request path. An unreachable registry hangs the first request indefinitely; concurrent first requests race the install with no lock.

Pre-install or lazy-load outside the request hot path, and give `installPlugin` a timeout.

apps/cli/src/proxy/buyer-proxy.ts
kotevcode, line 1: **`runVerifier` is called without the client abort signal or any timeout** (`buyer-proxy.ts:1220`)

The scope has `clientAbortController` (wired to client disconnect, passed to `_dispatchToPeer`), but it isn't forwarded to `runVerifier`, and `reach` passes no signal to `sendRequest`. A hung verifier plugin stalls the request indefinitely; a client disconnect doesn't cancel in-flight verification. (The seller-never-responds subcase is bounded by the 5-min request timeout, but the hung-plugin path is unbounded.)

Pass `clientAbortController.signal` through and wrap the verification in a timeout.

apps/cli/src/plugins/loader.ts
kotevcode, line 1: **Version-pin mismatch never self-heals, and the error advises a nonexistent command** (`loader.ts:56` / `:107`)

`ensureTrustedPluginInstallReady`'s staleness check ignores version, so a wrong-version install is never reinstalled; `assertPinnedPluginVersion` then hard-fails every load (seller `process.exit(1)`, buyer permanent "install/load failed"). The remedy text says `antseed plugin add <pkg>@<ver>`, but no `plugin` command is registered in the CLI. Separately, with `ANTSEED_SKIP_PLUGIN_UPDATE_CHECK=1` and the plugin absent, the error falsely claims "an unknown version is installed" instead of "not found."

Make the reinstall path version-aware, and fix the message (and the suggested recovery command).

apps/cli/src/cli/commands/buyer/start.ts
kotevcode, line 1: **Contradictory verifier flags are silently accepted** (`buyer/start.ts:408`)

`antseed buyer start --require-verifier --no-verifier` (verified against commander 14) parses to `{verifier:false, requireVerifier:true}`; `--no-verifier` wins, `verifierPolicy` becomes undefined, and the CLI prints "Verifier: disabled". An operator who explicitly demanded verification routes to unverified sellers with no warning. Same for `--verifiers X --no-verifier`.

Add a conflict guard that errors when `options.verifier === false && (options.requireVerifier || options.verifiers)` before line 408.

apps/cli/src/plugins/loader.ts
kotevcode, line 1: **Loader silently binds to any matching export instead of a designated entry point** (`loader.ts:75`)

The old strict `mod.default` + type check became a scan of `[default, ...Object.values(mod)]` that picks the first `type`+method match, ordered alphabetically by ESM export name. A package with a broken default but another matching export (mock/fixture/legacy) loads the wrong object with no diagnostic; provider/router paths have no name cross-check to catch it. Low practical severity today (all plugins trusted, single-export, pinned) but a real latent footgun.

Add a `candidates.length > 1` ambiguity error, or a name-vs-id check for provider/router like the prover/verifier paths already have.

apps/cli/src/plugins/registry.ts
essamhassan: done

apps/cli/src/cli/commands/buyer/start.ts
essamhassan: Fixed in 1beef17.

apps/cli/src/proxy/buyer-proxy.ts
essamhassan: Fixed in 1beef17.

essamhassan: Fixed in 1beef17

essamhassan: Fixed

apps/cli/src/cli/commands/network/browse.ts
essamhassan: As discussed offline I think this should be deferred to a later integration PR against your rewrite

apps/cli/src/plugins/loader.ts
essamhassan: fixed

packages/node/src/seller-request-handler.ts
essamhassan: Added a solution, please review

apps/cli/src/cli/commands/seller/start.ts
essamhassan: TRUSTED_VERIFIER_PLUGINS should be the way to go

apps/cli/src/plugins/verifier.ts
essamhassan: Fixed in `1beef17` verifiers preload at startup and the request path uses `loadVerifierPlugin(id, {install:false})`; `installPlugin` also got a 120s timeout + inflight dedupe.

apps/cli/src/cli/commands/buyer/start.ts
essamhassan: Verifiers come from the curated `TRUSTED_VERIFIER_PLUGINS` list (`1beef17`); default-on best-effort

PR #720, opened 2026-07-03
feat(verification): whitelisted verifier network with model fingerprinting

kotevcode:
## Summary

Adds an on-chain **verifier network** for model-identity verification — the "how do you know the seller is really serving the model it claims" problem from the [fingerprint-swarm blog post](apps/website/blog/2026-06-15-model-verification-fingerprint-swarm/index.md). Whitelisted verifiers act as buyers, probe sellers advertising a given model, compare response distributions across the cohort, and attest SAME/DIFF verdicts on-chain backed by ResponseAuth-signed evidence. Verifiers earn ANTS from the `VERIFICATION_MINTER_ID` emissions bucket.

> **Stacked on #632** (recognized-usage emissions). Base branch is `refactor/contracts-domain-architecture`; merge that first. The rewards contract is a gate bucket controller and the points policy plugs into `AntseedUsageAccounting`.

## How it works

1. A verifier runs `antseed verifier start` — **zero config needed**: the daemon is a full buyer node, so one wildcard discovery per round enumerates every service advertised in signed peer metadata and all of them become audit targets (`verifier.services` restricts the set when you want to).
2. Per service it picks a **stalest-first cohort** and commits a fresh private probe set on-chain *before* probing (commit-reveal — fabricated-after-the-fact results are rejected).
3. It probes every cohort member with identical requests over the ordinary paid buyer path, requiring a verified `ResponseAuth` on each.
4. **Cohort consensus** turns the majority behavior into the reference and flags statistical outliers — no trusted upstream API key needed. Certified KBF references strengthen/override the verdict and unlock auditing services with fewer than `cohortMinSize` sellers, down to a single seller.
5. It writes an evidence bundle and submits an attestation per seller. At epoch end it claims its share of the reward bucket pro rata to credited audits.

## Cohort consensus (why this beats single-seller KBF)

The blog framing — "test 10 kimi-k2 sellers, find the liar" — is stronger than reference-based KBF because it needs no trusted reference. When N sellers claim the same model, the majority answer distribution *is* the reference and outliers are the substitutes. Caveat: if an entire cohort serves the same substitute, consensus can't catch it — that's what KBF references (and future TEE references) are for.

## Reference enrollment

`antseed verifier reference build --model <id>` enrolls a model through any trusted OpenAI-compatible upstream (canonical provider, OpenRouter, or a local deployment of the open weights) following the published KBF protocol: candidate probes are consistency-filtered across temperatures, the median becomes the certified `consensus`, and an independent hold-out pass measures the reference's own error rate (`selfTest`) that Clopper–Pearson bounds p₀ from. With `verifier.upstream` configured (`baseUrl`, `apiKey`/`apiKeyEnv`, optional `modelMap`), the daemon enrolls references **automatically** for discovered services that lack one. References drift as backends update (~7–9 weeks); the daemon warns when one is stale.

## Stealth probes & probe hygiene

Probe requests are generated to read like **organic user chat** (varied + task-shaped framings, 1–3 facts per message, free-text numeric extraction) instead of a recognizable `TASK … output ONLY in (N) <number>` battery, so a cheating seller can't cheaply classify and special-case probe traffic. Tunable via `verifier.maxProbesPerRequest` (stealth vs. per-request-fee cost).

Probes come from a pluggable **`ProbeSource`**: a large compositional entity/attribute space by default (the static bank is a test fixture), with fresh per-audit selection and a per-service **rotation log** excluding probes already revealed by past audits — the spoofing literature shows a fixed public bank is fatal (a cheater can LoRA-finetune on it). All selection, nonces, and phrasing derive from one CSPRNG seed via RFC 5869 HKDF + SP 800-90A HMAC_DRBG; evidence hashing follows RFC 8785 (JCS) so committed artifacts reproduce byte-identically.

## Research grounding

The design matches the published KBF protocol ([arXiv:2605.29524](https://arxiv.org/abs/2605.29524)): 0/30 false positives and 60/60 true positives under deployment wrappers where LLMmap (37.5% FPR) and distribution tests broke; audits cost $0.002–$0.25 per endpoint. Related: [Model Equality Testing](https://arxiv.org/abs/2410.20247) (11/31 commercial Llama endpoints deviated from reference weights), [GhostPrint spoofing](https://arxiv.org/abs/2606.16100) (public probe banks are spoofable; diverse private distributions are the defense), and the known residual gap — near-lossless quantization of the true model is invisible to every black-box method; only TEE closes it (explicit non-goal here). Full citations in spec 07.

## Reputation & enforcement

- Registry accumulates per-(agent, service) and per-agent stats (same/diff/undetermined counts, distinct verifiers, last verdict) — facts only on-chain.
- Buyers compute a local authenticity score per (peer, model) during discovery (`PeerInfo.modelVerification`) and can route away from flagged sellers.
- `AntseedVerifierPointsPolicy` (swappable `IAntseedPointsPolicy`) zeroes/discounts a DIFF-flagged seller's recognized-usage points — a reversible economic penalty. Stake slashing still requires the arbiter dispute path (unchanged, still future work).

## Changes

**New package `@antseed/fingerprints`** — pure TS verifier math: KBF (Clopper-Pearson + one-sided binomial verdicts), cohort consensus, pluggable `ProbeSource` with a compositional generator, standards-based deterministic randomness (HKDF / HMAC_DRBG / Fisher–Yates with rejection sampling), RFC 8785 canonical evidence hashing, stealth request engine + free-text extractor. No P2P/chain deps.

**Contracts** (`packages/contracts/verification/`): `AntseedVerifierRegistry`, `AntseedVerifierRewards`, `AntseedVerifierPointsPolicy`, interfaces, `DeployVerifierNetwork.s.sol`, and `DeployRecognizedUsage.s.sol` wiring (verification bucket now points at the rewards contract).

**`@antseed/node`**: `VerifierRegistryClient`/`VerifierRewardsClient`, `serviceHash()`, `PeerInfo.modelVerification` + scoring, `getResponseAuth()`, chain-config + payments fields.

**`@antseed/cli`**: `antseed verifier start|status|claim` with service auto-discovery, per-epoch budget tracking, probe rotation log, and automatic reference enrollment; `antseed verifier reference build`; a `verifier` config section (incl. `upstream`, `probeSource`, `probeRotationHistory`); buyer `start` wires the verifier registry for reputation-aware discovery.

## Testing

- `@antseed/fingerprints`: 156 tests
- `@antseed/node`: 738 tests
- `@antseed/cli`: 208 tests
- `forge test`: 732 tests (baseline 660 + 72 new/updated), fuzz green at 10k runs
- Full `pnpm build` + `pnpm typecheck` clean

The rewards pro-rata split and the points policy's never-revert guarantee are fuzz-tested.

## Not in scope (future)

Verifier staking/slashing, non-KBF verifier families, the public fingerprint swarm, a managed buyer reference store, an off-chain exhibit re-verifier, on-chain seller stake slashing (current enforcement is the reversible points penalty), two-round re-query against mixed routing, anytime-valid sequential testing, and a passive idiosyncrasy tier-0 filter over organic completions.

## Deploy notes

Contract addresses land in `chain-config.ts` after deployment (deploy scripts print them). Post-deploy the owner must `setVerifier(...)` each approved verifier and may wire the points policy via `UsageAccounting.setPointsPolicy(...)`.


Augustas11 (commented):
Reviewed against the pinned diff at head `772576dc`. Two independent LLM reviewers, intersection of concerns only.

Architecturally this looks right. `AntseedVerifierRewards` freezes the epoch budget on first touch, uses `Math.mulDiv` for pro-rata splits, holds no ANTS, is `nonReentrant`, and the fuzz suite pins sum-of-claims ≤ frozen budget. The points policy's never-revert discipline is well-shaped. Deterministic randomness uses HKDF-SHA256 / HMAC_DRBG-SHA256 / Fisher–Yates rather than off-the-shelf gambles, and canonical JSON follows JCS. Two items I'd want to land before merge.

## 1. Single-verifier DIFF permanently zeroes a seller's emissions, and cohort consensus is not independence-constrained

**Where:** `packages/contracts/verification/AntseedVerifierPointsPolicy.sol` (around L92) with `packages/fingerprints/src/cohort.ts` (around L225).

The policy gates the penalty on `stats.diffCount > 0 || stats.lastVerdict == DIFF`. Both flags are monotonic — `diffCount` never decrements, `lastVerdict == DIFF` has no clearing path — and `diffPenaltyBps` defaults to 100%. Because `submitAttestation` does not require cohort agreement on-chain (each verifier submits independently), one attestation from any single approved verifier permanently denies the seller its share of the seller-emissions bucket. `test_diffFlagIsStickyAcrossLaterCleanVerdicts` codifies this exactly.

The off-chain reputation this is supposed to mirror already requires more evidence — `computeModelVerificationScore` needs `distinctVerifierCount >= 2` for a positive corroboration — so the on-chain penalty is strictly stronger than the routing signal it reflects. Cohort side compounds it: consensus in `cohort.ts` is an unweighted majority with no independence constraints (owner grouping, stake diversity, reputation weighting), so a coordinated cheap-substitute cohort can produce DIFF against an honest seller, and that single attestation permanently zeroes them.

**Suggested fix:** gate the penalty on `stats.distinctVerifierCount >= MIN_DISTINCT` and `stats.diffCount >= MIN_DIFF_ATTESTATIONS` (owner-configurable, defaulting to e.g. 2/2), drop `lastVerdict == DIFF` from the OR so a subsequent SAME can clear the flag, and refuse cohort-mode DIFF when `distinctVerifierCount` on that (agent, service) is 1.

## 2. `sampleRate: 1` gives sellers a per-peer verifier fingerprint

**Where:** `apps/cli/src/cli/commands/verifier/start.ts:388-390`.

The daemon forces `verification: { sampleRate: 1 }` and requests ResponseAuth on every request; ordinary buyers run `sampleRate < 1`. A seller building a per-peer histogram of "fraction of requests that ask for auth" clusters verifiers even under a fresh peer-id per round — a short-window estimator lands near 1.0 for probes and near the buyer default otherwise. Orthogonal to the content-layer stealth in `kbf/stealth.ts`: the seller doesn't need to detect stealth prompts to route probe traffic to the real model and substitutes to everyone else, which cuts against the indistinguishability invariant the docstrings call out.

**Suggested fix:** randomize per-request whether ResponseAuth is requested at a rate close to the buyer default; fail-close only if the fraction of unauth'd probes drops the verdict below the cohort's coverage threshold.

---

**Overall:** `COMMENT`. Direction is sound; item 1 is the one that would change my read from "merge-ready" to "merged with confidence" — currently exercisable by one approved verifier and permanent. Happy to take another pass once item 1 lands.

kotevcode:
Thanks for the review — item 1 was a real design flaw and is fixed in 56e2fda2; item 2 turns out to rest on a per-request auth flag that doesn't exist in the protocol. Details on both:

## 1. Fixed: corroborated, reversible DIFF penalty (56e2fda2)

Agreed on the substance, and it was arguably worse than stated: DIFF verdicts carry a designed false-positive rate (α per audit), so across enough honest audits a false DIFF is a statistical certainty — and the old gate (`diffCount > 0`, monotonic, no clearing path) turned that one unlucky audit into a permanent zero. It also contradicted spec 07's own "soft, reversible" framing.

One refinement over the suggested gate: `distinctVerifierCount >= 2 && diffCount >= 2` doesn't quite work, because `distinctVerifierCount` counts verifiers who attested *anything* — two SAME-attesting verifiers plus one DIFF would satisfy it. What landed instead:

- **Registry** now tracks `activeDiffVerifierCount`: the number of distinct verifiers whose *latest* verdict is DIFF. A verifier re-attesting SAME/UNDETERMINED on the same service retracts its own standing DIFF. This is a standing accusation, not a historical mark — `diffCount` stays monotonic as the audit-history record.
- **Agent-level standing** is any-service: a verifier's agent-level DIFF persists while *any* of the agent's services carries it (`_verifierDiffServiceCount`), so a seller cheating on kimi-k2 can't launder the flag by honestly serving a second model and collecting a SAME on it.
- **Policy** gates on `activeDiffVerifierCount >= minDistinctDiffVerifiers`, owner-configurable, default 2 — aligned with the routing score's corroboration bar. One mistaken/malicious/cohort-fed verifier can no longer zero a seller alone, and accusers retracting clears the penalty.

New tests: registry transition coverage (raise-once-per-verifier, retraction, re-raise, per-pair vs any-service agent semantics, monotonic `diffCount` preserved) and policy coverage (single-verifier DIFF passes through, corroborated DIFF zeroes, retraction restores, threshold 1 behavior, setter bounds/only-owner). `test_diffFlagIsStickyAcrossLaterCleanVerdicts` is replaced by `test_retractionRestoresSellerPoints`. 740/740 forge tests green, verification suites fuzz-clean at 10k runs. The TS client ABI/decoder picked up the new struct field.

On the cohort-independence point: agreed the consensus is unweighted, and that's exactly why the penalty threshold moved to ≥2 distinct verifiers — a colluding cohort feeding one verifier can no longer zero an honest seller. Independence constraints (owner grouping, stake diversity) inside cohort selection itself are worth doing as a follow-up; noting also that a certified-reference DIFF already overrides a passing cohort in `combineVerdicts`, which bounds the whole-cohort-collusion case in the other direction.

## 2. Not applicable as described: there is no per-request auth signal to histogram

The premise — "a seller building a per-peer histogram of 'fraction of requests that ask for auth'" — doesn't map to the wire protocol:

- ResponseAuth is **not requested per-request**. The seller emits it for *every* request whenever the buyer connection advertises the `response-auth.v1` capability — `seller-request-handler.ts:468-474` gates on `conn.hasRemoteCapability(CONNECTION_CAPABILITY_RESPONSE_AUTH_V1)`, a connection-level capability shared by every buyer running a current node.
- `verification.sampleRate` controls only whether the **buyer locally stores** the full request/response sample after verification — `VerificationSampler.shouldSample()` in `verification/samples.ts` is consulted on the buyer's disk-write path and never influences a byte on the wire.

So `sampleRate: 1` vs the buyer default of 0.02 produces identical traffic from the seller's viewpoint; the auth-fraction estimator would read 1.0 for every capable buyer on the network. The suggested fix (randomize per-request whether auth is requested) would require *adding* a per-request auth-request bit to the protocol — which would create exactly the fingerprint it's meant to hide.

The broader class of concern (transport-shape fingerprinting: fresh peer + new channel + a burst of short paid requests) is real and is called out in spec 07 as the residual gap of the stealth layer — the stated endgame there is scoring genuinely organic buyer traffic. Happy to expand that spec section if you think the threat deserves more than it currently gets.

PR #732, opened 2026-07-10
Validate seller payment metadata hash

kc-zero-lab:
## Summary
- reject paid seller SpendingAuth payloads when `keccak256(metadata)` does not match `metadataHash`
- apply the check before local recovery/reserve/update logic can store mismatched metadata
- update seller payment tests for mismatched and malformed metadata cases

## Why
`AntseedChannels._settleSpend()` recomputes `metadataHash = keccak256(metadata)` before verifying the buyer signature. Without the same seller-side check, a buyer could send a valid signature over hash A with raw metadata B, causing the seller to accept locally but later fail settlement/close/top-up with `InvalidSignature`.

## Tests
- `corepack pnpm --filter=@antseed/node test -- seller-payment-manager.test.ts`
- `corepack pnpm --filter=@antseed/api-adapter build && corepack pnpm --filter=@antseed/node typecheck`

### DIFF
--- a/packages/node/src/payments/seller-payment-manager.ts
+++ b/packages/node/src/payments/seller-payment-manager.ts
@@ -1,4 +1,4 @@
-import { type AbstractSigner, verifyTypedData } from 'ethers';
+import { type AbstractSigner, keccak256, verifyTypedData } from 'ethers';
 import type { Identity } from '../p2p/identity.js';
 import type { PaymentMux } from '../p2p/payment-mux.js';
 import type {
@@ -412,6 +412,14 @@ export class SellerPaymentManager {
       const cumulativeAmount = BigInt(payload.cumulativeAmount);
       const existingCumulative = this._acceptedCumulative.get(channelId);
 
+      if (!this._metadataMatchesHash(payload)) {
+        debugWarn(
+          `[SellerPayment] Rejecting SpendingAuth: metadataHash mismatch ` +
+          `channel=${channelId.slice(0, 18)}...`,
+        );
+        return 'rejected';
+      }
+
       const { channels: channelsAddr } = await this._resolvedAddresses!;
       const channelsDomain = makeChannelsDomain(this._config.chainId, channelsAddr);
 
@@ -678,6 +686,15 @@ export class SellerPaymentManager {
     } catch (err) {
       debugWarn(`[SellerPayment] Failed to process SpendingAuth: ${err instanceof Error ? err.message : err}`);
       return 'rejected';
+    }
+  }
+
+  private _metadataMatchesHash(payload: SpendingAuthPayload): boolean {
+    try {
+      return keccak256(payload.metadata).toLowerCase() === payload.metadataHash.toLowerCase();
+    } catch (err) {
+      debugWarn(`[SellerPayment] Invalid metadata payload: ${err instanceof Error ? err.message : err}`);
+      return false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Augustas11 (commented):
Nice fix — this closes a real gap. The buyer signs `SpendingAuth(channelId, cumulativeAmount, metadataHash)`, so only the *hash* is signed and the raw `metadata` blob rides alongside unsigned. Before this change the seller verified the signature but never checked `keccak256(metadata) == metadataHash`, so a buyer could attach a `metadata` whose hash differed from the signed one; the seller would accept and serve, then be unable to settle (the contract recomputes `keccak256(metadata)` and reverts on mismatch). Gating every path on the hash equality is the right invariant in the right place — before signature verification and before any RPC, so malformed payloads are rejected without a round-trip. Two small things before it lands.

## 1. Normalize empty/zero metadata so the check agrees with your own settle path

`_metadataMatchesHash` rejects `metadata === ''` (and omitted metadata): `keccak256('')` throws in ethers — `''` isn't a valid `BytesLike`, only `'0x'` is empty-bytes — so it's caught and returns `rejected`. Because the check runs before branch dispatch, this rejects on **every** path, including first-reserve.

The wrinkle is that the seller's own settle path still reconstructs the zero case:

```ts
// _getSettleParams
metadata: latestAuth.metadata || encodeMetadata(ZERO_METADATA),
```

`encodeMetadata(ZERO_METADATA)` hashes to `ZERO_METADATA_HASH` — exactly the hash a zero-metadata buyer signs — and the contract would settle that pair. So an auth with `metadata=''` + `metadataHash=ZERO_METADATA_HASH` is consistent from the contract's and the settle path's point of view, but the new check rejects it. It's stricter than both the contract and your own reconstruction, and the empty representation also shows up in hydration (`channel.latestMetadata ?? ''`).

Suggest normalizing before hashing, mirroring the settle fallback exactly:

```ts
const m = payload.metadata || encodeMetadata(ZERO_METADATA);
return keccak256(m).toLowerCase() === payload.metadataHash.toLowerCase();
```

This still rejects `''` paired with a *non-zero* hash (the actual attack) and still rejects malformed non-hex metadata — it only stops false-rejecting the zero-metadata compression the contract would have accepted.

You've clearly made this fail-closed on purpose — the renamed disconnect test flips `''` from "fall back to `0x`" to "reject before it can settle" — and that's a defensible call. So take this as flagging the *divergence*, not prescribing leniency: there are two consistent ways to resolve the auth-vs-settle asymmetry — normalize at the check (the one-liner above), **or** tighten `_getSettleParams` (the `|| encodeMetadata(ZERO_METADATA)` reconstruction) so both paths agree that `''` is invalid. Either aligns them. And to be clear on severity: the current buyer always sends `encodeMetadata(...)` on the wire (including `encodeMetadata(ZERO_METADATA)` for the zero case), so no shipped client hits the false-reject today — this is internal consistency/hardening, not a live break. Non-blocking.

## 2. One test worth adding, co-located with the new check

Credit where due — the accept path is already guarded: a `_metadataMatchesHash` that returned `false` unconditionally would fail `accepts SpendingAuth within deposit ceiling`, `recovers an active on-chain channel when local seller session is missing`, and `test_handleSpendingAuth_firstSign` (all build matching `metadata`/`metadataHash` via `buildSpendingAuth` → `computeMetadataHash`). So the regression net exists.

What's missing is a case named for and co-located with the new helper, plus the one interaction those don't isolate: **first-reserve with zero metadata** (`metadata = encodeMetadata(ZERO_METADATA)`, `metadataHash = ZERO_METADATA_HASH`) → `reserved`. That's exactly finding 1's surface, so it either catches the divergence or proves it moot. A malformed-non-hex-`metadata` → `rejected` case would round out the `try/catch` branch cheaply.

---

Verdict: **COMMENT** — the invariant is correct and worth landing; (1) is a divergence to reconcile either direction (non-blocking) and (2) is a small test add. Happy to re-review once those are in. Thanks for tightening the seller-side check to match the contract.

PR #734, opened 2026-07-12
fix(proxy): Preserve Responses stream mode

kotevcode:
## Description

Forces upstream OpenAI Responses requests to use `stream: true` while preserving the original client streaming preference through the buyer proxy and provider boundary. Also raises the default seller concurrency to 50 concurrent requests across CLI, Desktop config defaults, and the OpenAI Responses provider default.

## Release Notes

- Increased default seller concurrency from 5 to 50 concurrent requests.
- Fixed OpenAI Responses protocol transforms so non-stream clients receive JSON responses while upstream requests still satisfy the required streaming mode.

## Types of Changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project uses selective package publishing. Each release entry lists the pub
 - `@antseed/api-adapter`
 - `@antseed/cli`
 - `@antseed/node`
+- `@antseed/provider-openai-responses`
 
 ### Desktop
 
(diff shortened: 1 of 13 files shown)


kotevcode:
@claude

kotevcode:
Addressed the substantive review edge case around `Accept: text/event-stream` without a body `stream: true` flag. The buyer proxy now passes its effective streaming decision into the protocol adapter, so transformed upstream bodies, the internal Responses stream-preference header, and streaming response adapter creation stay aligned. Added adapter and proxy regression coverage for that case.

PR #740, opened 2026-07-15
fix(payments): Defer transient tx backpressure

kotevcode:
## Description

Treats temporary delegated-account transaction queue saturation as retryable seller payment backpressure. Top-ups are deferred instead of being classified as permanent failures, and close attempts preserve channel state so timeout cleanup can retry rather than dropping active buyer sessions.

## Release Notes

- Fixed seller payment handling so temporary delegated-account transaction queue backpressure defers top-ups and retries closes instead of permanently rejecting active buyer sessions.

## Types of Changes

## Checklist


kotevcode:
@claude

kotevcode:
Addressed the review feedback by adding dedicated zombie-close coverage for delegated-account transaction backpressure. The new test confirms repeated `checkTimeouts()` attempts keep the zombie channel active and continue retrying instead of incrementing toward local eviction.

The uncapped retry behavior for this specific transient queue-saturation error is intentional: local eviction is reserved for permanent close failures, while buyers still retain the on-chain `requestClose()` / `withdraw()` escape hatch.

Validation:
- `pnpm --filter @antseed/node test -- tests/seller-payment-manager.test.ts`

PR #745, opened 2026-07-21
fix(payments): stop seller close() retry loop after ReserveAuth mispersist

kotevcode:
## Summary

Fixes two related seller-side payment channel close bugs that together caused a permanent `close()` retry loop:

1. **ReserveAuth persisted as SpendingAuth** — when a channel opens, the buyer's initial signature is a ReserveAuth, but it was stored in the `latestSpendingAuthSig` column. After a seller restart, hydration restored it as if it were a SpendingAuth, so every `close()` attempt reverted with `InvalidSignature` (wrong EIP-712 type). The column is now left `null` at channel creation; only real SpendingAuths are persisted there.

2. **Infinite close retries** — when `close()` exhausted `MAX_CLOSE_RETRIES`, the channel remained `ACTIVE` in the store, so `checkTimeouts()` and every seller restart replayed the same doomed signed close forever. The channel is now persisted as `TIMEOUT` once retries are exhausted (both at the pre-check and when the final attempt fails), so it falls back to the timeout path. Intermediate failures also clean up `_activeBuyers` so the buyer session doesn't leak.

## Changes

- `packages/node/src/payments/seller-payment-manager.ts` — don't persist ReserveAuth sig as SpendingAuth; persist `TIMEOUT` status when close retries are exhausted; clean up `_activeBuyers` on intermediate close failures.
- `packages/node/tests/seller-payment-manager.test.ts` — new test covering retry exhaustion → persisted `TIMEOUT` → restart does not retry the signed close; updated the disconnect-close-failure test for the corrected persistence semantics.
- `CHANGELOG.md` — two entries under Unreleased → Fixed (user-facing seller behavior change in `@antseed/node`).

## Testing

- `vitest run tests/seller-payment-manager.test.ts` — 53/53 pass on this branch.
- `packages/node` builds clean (`tsc`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ### Fixed
 
+- Fixed the seller persisting the initial ReserveAuth signature in the SpendingAuth column, so restart hydration restored it under the wrong EIP-712 type and every subsequent `close()` reverted with `InvalidSignature`.
+- Fixed sellers retrying a failing signed `close()` forever: once the retry limit is exhausted the channel is now persisted as timed out, so `checkTimeouts()` and restarts fall back to the timeout path instead of replaying the same doomed close.
 - Fixed seller crashes when sending `PaymentRequired` to a buyer that disconnected before the payment terms could be delivered.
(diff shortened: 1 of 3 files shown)


kotevcode:
@claude

PR #746, opened 2026-07-25
feat(payments): Buyer-requested cooperative channel close

kotevcode:
## Description

Lets a buyer ask its seller to close a payment channel on the spot, releasing the reserved USDC without the on-chain `requestClose()` → 15-minute grace → `withdraw()` detour. Adds `CloseChannelRequest` (0x59) / `CloseChannelResult` (0x5A) to the existing PaymentMux, gated on the seller not being mid-accumulation with that buyer.

## How the seller decides

It agrees only when both hold:

- **No billable request in flight.** A new per-buyer in-flight counter spans the whole billable span — provider call, spend recording, and the follow-up `NeedAuth` — so a close cannot land between serving a request and claiming its cost. Otherwise `busy`.
- **No unsigned spend.** If `spent` exceeds the highest signed cumulative, the seller waits briefly for a catch-up auth already on the wire, then emits `NeedAuth` and rejects with `pending_auth` + `requiredCumulativeAmount` so the buyer signs and retries.

Rejections resolve normally rather than throwing, and leave the channel untouched — the buyer retries or falls back to the timeout path.

## Whose signature wins

The buyer may attach its latest SpendingAuth, and does by default. Attaching costs nothing — the cumulative is unchanged, so it authorizes no more than the seller could already claim — but it lets a seller that never received the last auth still close at the full amount owed. The seller settles at `max(own last-accepted, buyer-supplied)`, so neither side can settle below what is actually owed. A supplied auth must recover to the on-chain channel buyer and its `metadataHash` must match its `metadata`, else `invalid_auth`.

When neither side holds a usable auth, the seller closes at the current on-chain `settled` amount with an empty signature — the contract skips verification when `finalAmount == settled`, and this claims no unproven spend.

## Release Notes

- Added buyer-initiated cooperative channel close, so a buyer gets its reserved USDC released immediately instead of waiting out the on-chain `request-close` → 15-minute grace → `withdraw` flow. The seller closes on-chain and returns the transaction hash.
- The seller refuses while still mid-accumulation — a request in flight (`busy`) or work the buyer hasn't signed for (`pending_auth`) — leaving the channel untouched for a retry.
- Added `antseed buyer channels close <channelId>` (`--no-auth`, `--json`), which runs through a running `antseed buyer start` daemon's live seller connection.
- Added `AntseedNode.requestChannelClose(peerId, opts)` and the `payments.cooperative-close.v1` capability.

## Types of Changes

## Checklist

- [x] I have added the necessary documentation (payments spec §1 + §11, CHANGELOG)
- [x] I have added tests (13 seller-close, 6 negotiator round-trip, 6 codec)
- [x] Lint and unit tests pass locally with my changes (801/801 node, 201/201 CLI)

## Note

This also corrects the stale P2P message table in `docs/protocol/spec/04-payments.md`, which listed `0x50` as `ReserveAuth` and `0x54` as `SpendingAuth` — neither has matched `protocol.ts` for some time. Happy to split that out if you'd rather keep this PR to the feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ## Unreleased
 
+### Added
+
+- Added buyer-initiated cooperative channel close (`CloseChannelRequest`/`CloseChannelResult`, message types 0x59/0x5A), so a buyer can get its reserved USDC released immediately instead of waiting out the on-chain `request-close` → 15-minute grace → `withdraw` flow. The seller closes on-chain and returns the transaction hash. It refuses while it is still mid-accumulation with that buyer — a billable request in flight (`busy`), or served work the buyer hasn't signed for yet (`pending_auth`, sent alongside a `NeedAuth` for the outstanding amount) — leaving the channel untouched so the buyer can retry or fall back to the timeout path.
+- The buyer attaches its latest SpendingAuth by default; the seller closes at whichever cumulative is higher (its own or the buyer's), so a seller that lost the last authorization can still be paid in full, and a buyer cannot use this path to settle below what it owes.
+- Added `antseed buyer channels close <channelId>` (with `--no-auth` and `--json`), which runs the request through a running `antseed buyer start` daemon's live seller connection via the new `/_antseed/channels/close` control-plane endpoint.
+- Added `AntseedNode.requestChannelClose(peerId, opts)` to `@antseed/node`, plus the `payments.cooperative-close.v1` capability advertised in discovery metadata and the connection handshake.
+
 ### Fixed
 
 - Fixed the seller persisting the initial ReserveAuth signature in the SpendingAuth column, so restart hydration restored it under the wrong EIP-712 type and every subsequent `close()` reverted with `InvalidSignature`.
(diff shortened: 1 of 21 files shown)


kotevcode:
@claude

kotevcode:
Thanks — the finding is correct, and digging into it turned up something that changes the fix.

`peerSupportsCooperativeClose()` was indeed dead code, so there was no fast-fail path. Fixed in 6ec5b8b.

**One caveat worth flagging:** the obvious fix — gating on `conn.hasRemoteCapability(...)`, which is how `seller-request-handler.ts:562` checks response-auth — would have broken *every* cooperative close. `setRemoteCapabilities()` is only called from the two inbound-connection paths (`connection-manager.ts:746,772`, both `isInitiator: false`), so on the buyer's own **outbound** connection the remote-capability set is always empty. Gating on it would reject 100% of closes.

The working source is the peer's discovery metadata via `_peerCapabilities`, which is what `broadcastSweepRequest` already uses for `peerRelaysSweeps`. That's the precedent I followed.

Also made the check fail fast only on *positive* evidence — a peer that advertised some capabilities but not this one. An empty/unknown set means we never learned them (partial discovery data), so it attempts anyway rather than blocking a close that would have worked. An older seller still advertises `verification.response-auth.v1`, so it lands in the fast-fail branch correctly.

Added `tests/cooperative-close-capability.test.ts` covering all three cases, and documented the inbound-only pitfall in the spec so the next person doesn't reach for `hasRemoteCapability`.

Node 804/804, CLI 201/201.

PR #750, opened 2026-07-26
feat(buyer): Cool down unresponsive peers and fail over

kotevcode:
## Description

When an automatically selected peer stopped responding, nothing happened. The proxy recorded a diagnostic-only failure and returned 502, the desktop classified that as retryable and retried the **same dead peer twice**, and the next chat re-ran a scorer with no liveness term — so it picked the same dead peer again, indefinitely. The only blacklist was the manual `blockedPeerIds` list.

Peers that fail now enter a temporary cooldown, automatic routing deprioritizes them, and an auto-routed chat whose peer dies retries on a different healthy peer. Cooldowns escalate 30s → 8min and are cleared by any response; there is no permanent auto-ban, since peers go offline for legitimate reasons.

## Not blaming peers for our own faults

This is most of the work. Nearly every failure reaches one catch where a dead seller is indistinguishable from a dead chain RPC, an empty wallet, or a closed local socket — and a naive cooldown would have made things actively worse. There is no suspend handling anywhere in the repo and every timeout is `Date.now()` + `setTimeout`, so **one closed laptop lid** fires every pending timeout and keepalive at once and would have exiled the entire peer set. Cooldowns persist, so it would have survived the restart too. No amount of error-string matching catches that, or a dropped wifi connection.

So attribution is evidence-based and fails safe. Three independent gates in `peer-attribution.ts`, each sufficient alone:

- **Corroborating success** — a failure escalates only if some *other* peer succeeded recently, which is the cleanest available proof that our network, DHT, RPC and wallet are all fine. Two properties fall out: cold start penalizes nothing, and a buyer with a single peer never cools it down — correct, since there is nothing to fail over to.
- **Correlated-failure suppression** — 3 distinct peers failing within 60s means the common factor is us. Suppresses, and **rolls back** the streaks that outage already recorded.
- **Suspend detection** — a heartbeat watching for wall-clock jumps, rolling back everything since the last trustworthy tick.

On top of that, errors now carry a `buyer | peer | unknown` attribution from the site that raises them (`AntseedRequestError`, `faultAttributionOf`). Untagged stays `unknown` — blaming a peer requires evidence, so the default is never `peer`. Buyer faults return **503, not 502**, and are classified non-retryable, so an empty deposit shows the actual fix rather than silently walking the entire peer list.

## Two bugs found on the way

- **An unreachable chain RPC was blaming innocent peers.** A failed `getBuyerBalance` was swallowed into a `debugWarn` and negotiation continued blind, stalling for the full 30-second lock-confirmation timeout before failing with a 502 that blamed the seller and leaked the internal "Likely causes" prose to the client. Now an immediate 503 `chain_rpc_unavailable`, matching the `deposits_not_configured` path directly alongside it. Independently useful.
- **`SellerAuthorizationError` meant two opposite things** — "peer is not an authorized operator" (theirs) and "isOperator RPC failed" (ours) — distinguishable only by matching the message text.

## Why cooldown can never break routing

Advisory at three independent layers: the proxy still dispatches to a cooling-down peer when a request names it, `isPeerRoutable` stays block/allow-list only, and peer exclusion falls back to the unfiltered candidate set when it would empty the list. An all-peers-cooling-down deadlock is impossible by construction. Conversations pinned to a peer by hand are never moved off it.

One related fix: the success-clear widens from 2xx/3xx to any response below 500 (bar 408). A peer answering 400 to everything is still reachable, and requiring 2xx would strand it in a cooldown it could never escape.

## Release Notes

- Added buyer peer health cooldowns, so a seller that stops responding is temporarily deprioritized by automatic routing instead of being selected again on every request. Cooldowns escalate from 30 seconds to a maximum of 8 minutes, are cleared by any response from the peer, and are advisory only — a pinned or explicitly named peer is always still dispatched to. Exposed as `GET /_antseed/peer-health` and `POST /_antseed/peer-health/clear`.
- Added automatic peer failover for Desktop chats using automatic routing: when the bound peer stops responding, the retry moves to the next-best healthy peer and the chat reports which peer it switched to. Chats pinned to a peer by hand always keep that peer.
- Added fault attribution to request failures, so buyer-side problems such as empty deposits, an unreachable chain RPC, or a closed local transport are reported as buyer faults with HTTP 503 instead of peer failures with HTTP 502.
- Fixed buyer payment negotiation so an unreachable chain RPC is reported immediately as `chain_rpc_unavailable` rather than stalling for 30 seconds and then failing with an error that blamed the seller.
- Fixed `SellerAuthorizationError` so an unauthorized peer is distinguishable from an unreachable chain RPC.

## Types of Changes

## Checklist

- [x] I have added the necessary documentation (CHANGELOG)
- [x] I have added tests (15 attribution-gate, 18 peer-health state, 11 buyer-proxy incl. corroboration/rollback/control-plane, 9 fault-attribution, 8 VPR scoring, 4 route projection, 5 chat failover, 5 peer-cache, 4 route-mode binding, 3 stream classification)
- [x] Lint and unit tests pass locally with my changes (783/783 node, 295/295 CLI, 175/175 desktop, 65/65 e2e; full `pnpm run typecheck` and `pnpm run build` clean)

## Notes for review

- The escalation curve already existed in `router-core`'s `PeerMetricsTracker` but was unreachable dead code, only callable through the orphaned `LocalRouter.selectPeer`. It is now single-sourced as `computeFailureCooldownMs` and the tracker calls it; behaviour is unchanged.
- One e2e test needed updating: `payment-negotiation.test.ts` points the buyer at an unreachable RPC and asserted 402. It now gets the new 503, so the assertion was widened rather than weakened.
- Deliberately left out of scope: a "cooling down" badge on Discover cards, and a "try again now" button for the `peer-health/clear` endpoint (endpoint exists and is tested, no UI yet).


kotevcode:
@claude

kotevcode:
@claude

apps/cli/src/proxy/buyer-proxy.ts
```diff
@@ -1823,11 +2054,12 @@ export class BuyerProxy {
           })
         }
 
+        this._recordPeerResponseHealth(
+          selectedPeer.peerId, responseForClient.statusCode, requestForPeer.path,
```
kotevcode: `responseForClient` can be a buyer-generated negotiation response, not a response from the seller. In particular, `handle402()` now synthesizes 503 for `chain_rpc_unavailable` and `deposits_not_configured`; this call classifies either as `seller-5xx`, and after corroboration can increment/cool down an innocent peer. The router result immediately above is also recorded as a peer failure. Please propagate attribution/code alongside synthesized responses (or inspect the structured body here) so buyer-side 503s do not affect peer health/metrics.

packages/node/src/p2p/connection-manager.ts
```diff
@@ -201,7 +202,7 @@ export class PeerConnection extends EventEmitter {
   /** Send a message through the active transport. */
   send(data: Uint8Array): void {
     if (this._state !== ConnectionState.Open && this._state !== ConnectionState.Authenticated) {
-      throw new Error(`Cannot send in state ${this._state}`);
+      throw buyerFault(`Cannot send in state ${this._state}`, "buyer-transport-closed");
```
kotevcode: A closed connection is not necessarily a buyer fault: this state is also reached when the remote peer disconnects or the connection dies between acquisition and `send()`. Tagging it `buyer-transport-closed` makes the proxy return structured 503 and Desktop treats it as non-retryable, so the exact dead-peer case this PR is meant to fail over from can stop without failover. This should remain `unknown` (letting the attribution gates decide), or be tagged `peer` only where remote closure is proven; reserve `buyer` for an explicitly local shutdown/abort. The no-writable-transport branch below has the same issue.

kotevcode (commented):
Found two fault-attribution paths that undermine failover and cooldown behavior.

PR #761, opened 2026-08-03
feat(seller): Model health checks with auto-unadvertise

kotevcode:
## Description

Sellers now verify their advertised models actually work, and buyers react correctly when a model is gone.

- **Seller self-checks** (`ModelHealthChecker` in `@antseed/node`): every 5 minutes (configurable), each advertised service gets a 1-token probe through the provider relay — validating the exact path buyers hit, including auth and service-alias rewrites. After 3 consecutive failures the service is removed from the advertised list and signed metadata is refreshed via the new `AntseedNode.refreshSellerMetadata()`; removed services keep being probed and are re-advertised on recovery. Probe classification is three-way (healthy / unhealthy / inconclusive) so rate limits or probe-parameter rejections never unadvertise a working model, and a provider's last service is never removed (an empty services list means wildcard in metadata semantics).
- **Metadata capability**: sellers running self-checks advertise `seller.model-health.v1` in discovery metadata.
- **CLI**: new `seller.healthCheck` config (`enabled`, `intervalMs`, `failureThreshold`), enabled by default. When an ant-agent wrapper is configured, probes go through the unwrapped provider so they don't run full agent loops.
- **Buyer**: `model_not_found` responses now count as routing failures (previously routing successes) and trigger a throttled background refresh of peer metadata, so a stale cached model list recovers quickly after a seller unadvertises a model. Control-plane `/v1/models` lookups are excluded.

## Testing

- 15 new vitest cases for `ModelHealthChecker` (removal/recovery thresholds, inconclusive statuses, wildcard guard, timeouts, capacity skip, probe shapes per API protocol)
- 7 new CLI tests (config load/validation passthrough, `model_not_found` detection)
- Full workspace typecheck and `packages/node` (825) + `apps/cli` (205) suites pass

## Changelog

`CHANGELOG.md` updated under Unreleased (user-facing seller + buyer behavior change).


kotevcode:
@claude

PR #777, opened 2026-08-04
fix: Connect WSL installs of CLI tools and buyer state temp-file leak

kotevcode:
## Description

Two independent fixes, both user-facing:

### Desktop: config-patch connects now reach WSL installs and fail loudly when the tool is absent

The connect buttons for OpenCode, Codex, Crush, Goose, and pi wrote each tool's config only at its Windows-side path, which a tool living inside WSL never reads — and the patcher created config directories even when the tool wasn't installed anywhere, so the row flipped to "Connected" having connected nothing.

Connecting one of these tools now:

- **Detects WSL distros carrying an install** (`wsl.exe -l -q`, probing PATH plus per-tool home locations like `~/.opencode/bin`, `~/.codex`, `~/.local/bin`, and the tool's config/state dirs — see `WSL_TOOL_PROBES`) and patches the config inside each distro through the `\\wsl.localhost\` share. For Crush and Goose, whose native `configPath` is resolved to a Windows known folder at definition time, the profile carries a `wslConfigPath` with the XDG path distro-side installs actually read.
- **Writes a base URL the distro can actually reach**: `localhost` under mirrored networking; under default NAT networking the distro's default-route gateway (the Windows host's vEthernet address), where the desktop now runs a small TCP relay to the loopback-only buyer proxy. The buyer proxy stays bound to 127.0.0.1 on purpose — it is keyless, so exposing it beyond loopback would let anyone on the LAN spend the buyer's funds; the relay listens only on the WSL-facing address.
- **Persists the applied WSL targets per tool** (`system-proxy.wsl-targets.json`) so disconnect/quit unpatches the same files and stops the relay without re-running discovery, which would cold-start distros on the quit path.
- **Throws a descriptive error when the tool is found neither natively nor in any WSL distro**, instead of fabricating a config directory and reporting success. Native install signals: the config dir, the tool's home install/state locations, or the binary on PATH — any one suffices.

The new behavior is opted into per profile via an `installProbe` flag on the patch definition; generic profiles (packaged/env overrides, custom formats without the flag) keep the previous single-path behavior. Zed and T3 Code are native Windows GUI apps with no WSL install to patch, so they are unchanged.

### CLI: buyer state temp-file leak and silently dropped state writes

`buyer.state.json` is written atomically via a `.buyer.state.<uuid>.json.tmp` sibling plus `rename()`. When the rename failed — common on Windows while a reader briefly holds the state file open — the catch block swallowed the error, leaving the temp file behind (each one a full discovered-peers snapshot, ~1 MB) and silently dropping the state update: session pin, default route, and peer-cache changes were simply lost.

Now:

- The rename is **retried with a short backoff** for the transient Windows sharing-violation codes (`EPERM`/`EACCES`/`EBUSY`), recovering the write instead of dropping it.
- A write that still fails **unlinks its temp file and logs the error** (state writes are minutes apart, so this is audible without being noisy).
- Startup **sweeps leftover temp files** older than a minute, cleaning up what earlier runs left behind; the age floor protects a temp file another process (`antseed buyer connection set`) may be writing concurrently.

## Changelog

`CHANGELOG.md` updated under Unreleased → Fixed for both changes.

## Testing

- `apps/cli`: 279/279 (`node --test`), including new coverage for the merge-write temp-file cleanup, the startup sweep age/pattern rules, and missing-directory tolerance.
- `apps/desktop` main: 189/189 (`node --test`), including new tests for the WSL distro-list/gateway parsers, UNC path mapping, per-tool probe commands, per-tool targets-file round-trip and isolation, base-URL host substitution, install-probe applies (opencode, codex), and WSL unpatch-on-remove (opencode, goose — the latter covering the non-loopback gateway-host ownership check).
- WSL discovery and the relay bind are Windows-runtime paths that can't execute on macOS/Linux; the pure logic around them (parsers, path mapping, persistence, host substitution) is what the tests pin down. Worth a manual pass on a Windows machine with a WSL-side tool before the next desktop release.


kotevcode:
@claude

kotevcode:
Both review points addressed in 2f2a64e7:

**Per-profile failure isolation** — each `applyConfigPatch` in the `startSystemProxyRuntimeInner` loop now runs in its own try/catch. A failing profile is logged, dropped from `activeProfileNames` (so its row doesn't claim Connected), and its stale WSL target rows are cleared so the relay doesn't outlive the connection. Launch restore and union re-applies keep every still-working tool connected. The loud-failure behavior is preserved where it matters: if the failed profile is one this call was asked to *add* (i.e. the tool the user just clicked Connect on), the error is rethrown after the survivors are fully wired and persisted, so the UI still surfaces it. This distinction works because `loadPersistedSystemProxyState()` runs before the launch restore, so restored profiles land in `previousProfiles` rather than `newlyConnectedProfiles`.

**Malformed UNC guard** — `discoverWslToolTargets` now refuses config paths that aren't `~/`-rooted or absolute POSIX (`isMappablePosixPath`), so a custom crush/goose profile with `installProbe` but no `wslConfigPath` skips WSL patching instead of concatenating a Windows known-folder path onto `\\wsl.localhost\<distro>`. Native patching and the not-found error still apply for such profiles.

Desktop main tests: 190/190 (added coverage for the path guard, including the crush known-folder fallback case). The isolation logic itself lives in `runtime.ts`, which imports electron and isn't reachable from `node --test` — covered by reasoning and the manual Windows pass already planned before release.

PR #779, opened 2026-08-05
feat(desktop): Per-chat seller pinning and routed-peer visibility

kotevcode:
## Description

Chats currently pin themselves to the first peer that serves them, and nothing the user changes afterwards can move them: pinning a seller for a model only affects new chats, and re-pinning a chat resolves through auto-selection, ignoring the model's seller pin. This PR makes peer routing for existing chats controllable and visible.

The rule implemented: **deliberate actions move chats; ambient re-ranking never does.** A chat's remembered peer is affinity (kept across reputation shifts and new peers, preserving prompt cache), but an explicit seller pin outranks it.

## Changes

**CLI (buyer)**
- Conversation records track how the pinned peer was chosen (`peerSource: 'auto' | 'user'`); the conversations update endpoint accepts the flag. Older `conversations.json` files heal to `'auto'`.

**Desktop — per-chat seller control**
- Each model row in a chat's detail page has a settings button opening per-chat model settings: an "Auto select seller" toggle plus a seller list (reputation-sorted, `• Auto`/`Pinned` badges on the active route). A picked seller is stored as `peerSource: 'user'` and nothing moves it until released; turning auto back on re-resolves the route rather than keeping the previous pick.
- Pinning a seller for a model (Models page) re-points that model's existing auto-routed chats to the chosen seller; user-pinned chats stay put.
- Chat re-pinning resolves through the model's seller pin before falling back to the preference-scored best route, and re-clicking a chat's current model no longer resets its peer.
- The chat detail's model list drops the discount column and moves the settings button onto the title line, giving the meta line (peer name, seller count, price) the full row width.

**Desktop — routed-peer visibility**
- New "Show routed peer" preference (Preferences → Floating window, default off): chat rows name the seller each chat's requests actually went to next to the model — on the floating pill, the Chats page list, and the Home recent-chats card. Resolved against the unfiltered discover list so requests landing on excluded peers are visible too.

## Changelog

Updated `CHANGELOG.md` (Unreleased → Added) for both user-facing changes.

## Testing

- CLI: full test suite passes (280 tests), including new `peerSource` lifecycle coverage.
- Desktop: renderer + main typecheck clean, renderer test suite passes (171 tests), Vite renderer build succeeds.


kotevcode:
@claude

PR #793, opened 2026-08-07
feat(desktop): Outerfound card deposit via antseed-pay Stripe checkout

kotevcode:
## Summary

Integrates the AntSeed Pay hosted checkout (antseed-pay.com, Stripe-powered — merchant: Outerfound, Inc.) as a first-class desktop deposit path for US users. Pairs with refoundhq/antseed-pay#3 on the pay-page side.

- **Region-gated availability**: the main process asks the pay page's public `GET /api/options` which providers serve this machine's region (geo resolved from the request IP at Cloudflare's edge — no locale guessing). Cached per app run on success; **fails closed** on any error, so an unreachable page or unsupported region leaves the chooser exactly as it is today.
- **Chooser layout**: when Stripe is available, the primary CTA becomes "Deposit — Powered by Outerfound" with Visa/Mastercard/Amex badges (Stripe's onramp takes cards + US bank via Link — no Apple/Google Pay, so no wallet badges), and the Fun checkout moves under "More options" as a fully-functional "Deposit using Fun" row. Non-US users see the existing layout unchanged.
- **Checkout window**: selecting it opens a narrow (420×700) app-owned popup via the shared checkout-window infrastructure — Chrome UA, recursive popup adoption, and auto-close when the deposit watcher sees the purchased USDC land. The funding link is signed as before and now carries `provider=stripe`, which the new pay page renders as a single-provider checkout.
- **IPC surface**: new `payments:onramp-availability` handler + `paymentsOnrampAvailability` bridge; `payments:open-card-provider` routes antseed-pay to the popup (no wallet extension needed — the link is pre-signed) and logs the minted funding link in dev.
- CHANGELOG updated (user-facing desktop change).

## Test plan
- `tsc` clean for both main and renderer configs; all 190 main-process tests and 171 renderer tests pass.
- Manually verified end-to-end against the pay page dev server (test-plane Stripe keys): row gating via `/api/options`, popup opening on the signed `provider=stripe` link, live quote + Stripe embedded form, USDC delivery path unchanged (hot-wallet watcher).
- Fail-closed verified: with the pay page unreachable, the Stripe row is absent and the chooser matches current production behavior.

## Notes
- End-to-end against production activates once refoundhq/antseed-pay#3 is deployed (the live worker predates `/api/options`, so until then the row simply stays hidden).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ### Added
 
+- Desktop: added a card deposit path powered by AntSeed Pay (antseed-pay.com) with Stripe checkout for US users. The app asks the hosted pay page which providers serve the user's region (its public `/api/options` endpoint, geo-resolved at Cloudflare's edge) and, when Stripe is available, leads the Add Credits chooser with a "Deposit — Powered by Outerfound" CTA (Visa, Mastercard, Amex); the Fun checkout moves under "More options" as "Deposit using Fun". Selecting it opens a narrow app-owned checkout window (420×700) on the signed funding link pinned to the Stripe integration (`provider=stripe`); the existing deposit watcher sweeps the purchased USDC into credits and closes the window when funds arrive. The region check fails closed — if the pay page is unreachable or the region unsupported, the chooser is unchanged.
+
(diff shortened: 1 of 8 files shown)


kotevcode:
@claude

kotevcode:
1a5a17cb: funding-link log is now gated behind `isDev` — production console no longer prints the buyer address/signature.

On the minor notes: the brief Fun-CTA flash is accepted as the fail-closed trade-off — availability is cached per app run after the first success, so it only occurs on first mount; and test coverage for `portal.ts`/IPC follows the existing pattern (none today), left for a separate pass.

PR #796, opened 2026-08-07
refactor(node): extract @antseed/protocol and @antseed/buyer-core

kotevcode:
## Description

Splits two portable packages out of `@antseed/node`, with zero behavior change:

- **`@antseed/protocol`** — the wire protocol: message types and payload interfaces, the 9-byte frame codec, the binary HTTP request/response codec, payment JSON codecs, EIP-712 payment signatures (SpendingAuth/ReserveAuth, metadata v2, `computeChannelId`), the signed `intro`/`hello` connection-auth envelope, EIP-191 signing helpers, `PeerId`, peer metadata/capability/service-api types, and `ConnectionState`. Depends only on ethers; one subpath export per module.
- **`@antseed/buyer-core`** — the transport-agnostic buyer machinery: `BuyerRequestHandler`, `BuyerPaymentManager`, `BuyerPaymentNegotiator`, `BuyerFreeUsageManager`, the `ProxyMux`/`PaymentMux`/`VerificationMux` frame multiplexers, pricing and channel accounting, and the read-only Deposits/Channels EVM clients. Depends on `@antseed/protocol`, `@antseed/api-adapter`, ethers, and tokenx.

## Why

These modules were already logically independent of the node runtime but were entangled with it at the file level, so any other consumer of the protocol or buyer flow had to either depend on the full native-module-bearing `@antseed/node` or duplicate code. The split makes the boundary explicit and enforced by the package graph: neither new package can reach sqlite, sockets, or other Node-only APIs.

## How

- **Verbatim moves.** All moved files keep their code; only import paths changed. Every original `@antseed/node` path remains as a one-line re-export shim, so the node's public API and all internal imports are unchanged — the CLI, desktop, and plugins needed no edits.
- **Structural interfaces instead of node types.** The buyer machinery's environment coupling turned out to be tiny and is now expressed as interfaces in `buyer-core`: `BuyerConnection` (`state`, `stateChange` events, `send`), `BuyerChannelStore` (the channel bookkeeping surface), `BuyerIdentity`, `BuyerPeerView`, and optional verification sinks. The existing `PeerConnection` and sqlite `ChannelStore` satisfy them structurally; the only hand edits outside shims are three cast-annotated wiring lambdas in `node.ts`.
- **Two mechanical substitutions** in moved code: `node:crypto` `randomBytes` became ethers `randomBytes` (byte-identical salts), and `debug.ts` reads env vars through a `typeof process` guard.
- Build order: `protocol` → `api-adapter` → `buyer-core` → `node` (root tier0 updated). Publishing `@antseed/node` will require publishing the two new packages first.

One naming collision made explicit: `signatures` (SpendingAuth metadata, `2n`) and `peer-metadata` (discovery records, `10`) both define `METADATA_VERSION`. The `@antseed/protocol` root export resolves to the payments constant, matching `@antseed/node`'s existing public API; the discovery constant lives at the `./peer-metadata` subpath.

## Verification

- `@antseed/node`: 839/839 tests pass unchanged.
- `@antseed/cli`: 280/280 tests pass unchanged.
- Full workspace typecheck and tiered build (node → provider-core/router-core → plugins → cli) pass.

## Changelog

No user-facing behavior change — internal restructuring only, so no `CHANGELOG.md` entry.


kotevcode:
@claude

PR #687, opened 2026-06-16
Add OpenAI image generation support end-to-end

alexanderludwig:
## Summary
- add `openai-images` support to protocol detection and service API typing
- extend the buyer proxy and OpenAI provider to handle `/v1/images/generations`
- add provider, SDK, and end-to-end coverage for OpenAI image generation, including a local-chain flow script

## Why
AntSeed already supports OpenAI-compatible chat and responses flows, but image generation requests were not yet wired through the same adapter, proxy, and payment path. This change brings `/v1/images/generations` into that flow so image requests can be routed, paid for, and validated end-to-end.

## Validation
- ran `pnpm --filter @antseed/api-adapter test -- tests/adapter.test.ts`
- ran `pnpm --filter @antseed/node test -- tests/service-api-adapter.test.ts`
- ran `pnpm --filter @antseed/provider-openai test -- src/index.test.ts`
- ran `pnpm --filter @antseed/e2e test -- tests/openai-images-sdk.test.ts`
- ran `pnpm --filter @antseed/e2e test -- tests/openai-images-payment-flow.test.ts`

PR #800, opened 2026-08-09
Add OpenAI image generation support end-to-end

kotevcode:
Replaces #687, whose PR object got stuck server-side (head frozen at an old commit; pushes to `add-image-gen-support` stopped generating sync/CI events). Same content, plus #788 and review fixes. Also includes #788 (per-service model capabilities), which was merged into this branch.

## Summary
- add `openai-images` support to protocol detection and service API typing
- extend the buyer proxy and OpenAI provider to handle `/v1/images/generations` and `/v1/images/edits`
- per-unit image billing (`output_images` components with size/quality matching) announced in discovery metadata and verified on both sides of the payment flow
- announce per-service model capabilities (context window, max output tokens, input modalities, reasoning/tool-use/structured-output) via `ANTSEED_SERVICE_CAPABILITIES_JSON`, carried in metadata v12 (#788)

## Review fixes on top
- payment integrity: positive unit-billing claims (`NeedAuth`) are rejected until the buyer has observed the delivered response, so a seller cannot get paid for undelivered images; buyer waits briefly for its own response processing to close the race
- wire format: capabilities moved to a dedicated metadata v12 instead of redefining v11 in place; code-unit sort in signed bytes, dedicated value bits, entry-count guard, unknown-bit rejection
- buyer free-service gate resolves the same provider+protocol billing route the seller uses (fixes zero-token-priced image services being treated as free)
- capability validation shared between config parsing and the metadata validator (bad config fails fast at startup instead of silently killing announcements)

## Changelog
`CHANGELOG.md` updated (Added: image generation, service capabilities; Fixed: free-gate classification, NeedAuth fail-open).

## Validation
- full workspace build + typecheck clean
- `@antseed/node` 879/879, `@antseed/provider-core` 48/48
- full e2e suite 68/68, including `openai-images-payment-flow`


kotevcode:
@claude

PR #802, opened 2026-08-09
Add output modalities and supported-parameter hints to service capabilities

kotevcode:
Closes the two capability-discovery gaps left after #800: buyers had no way to learn what a service *produces* (image output was only implied by the `openai-images` protocol) or which optional request parameters it honors.

## Summary
- add `outputs` (modalities: text/image/audio/video/pdf) and `supportedParameters` (lowercase snake_case request-body parameter names, max 32) to `ServiceCapabilities`
- encode both in metadata v12 using the two free presence bits (6 and 7) — no version bump needed since v12 is unpublished; supported parameters are announced in code-unit sorted order so re-encoding decoded metadata stays byte-stable for signature verification
- validation is shared between seller config parsing and buyer-side metadata validation via `validateServiceCapabilityFields`, so anything a seller can configure is guaranteed to announce
- the `openai` provider automatically advertises `outputs: ["image"]` for `openai-images` services; explicit capability config extends or overrides the default per field
- the buyer proxy logs an advisory warning when an outgoing request carries body parameters the routed peer did not announce for that service (JSON and multipart edits bodies both checked); requests are never rejected, protocol-core fields are always allowed, peers announcing no list are never flagged, and transformed routes skip the check
- rename `SERVICE_CAPABILITY_INPUTS`/`ServiceCapabilityInput` to `SERVICE_CAPABILITY_MODALITIES`/`ServiceCapabilityModality` since the vocabulary now serves both directions (safe: introduced in #800, unpublished)

Capabilities stay hints, not enforced contracts: a chat-protocol service can legitimately announce `outputs: ["image"]` (e.g. OpenRouter's chat-shaped image models), so no protocol/output cross-validation is imposed.

## Testing
- codec roundtrip incl. re-encode byte-stability, validator accept/reject cases, config parsing, plugin default + override, and 6 new buyer-proxy parameter-check tests
- full workspace typecheck and build; node 885, provider-core 50, protocol 18, provider-openai 14, cli 290 tests passing

## Changelog
`CHANGELOG.md` updated (Unreleased → Added). Docs updated: configuration guide, discovery protocol page, wire-format spec, metadata v12 upgrade guide.


kotevcode:
@claude

PR #804, opened 2026-08-10
Guided protocol-aware capability prompts in seller setup

kotevcode:
Stacked on #802 (base is `feat/capability-outputs`; retargets to `main` automatically when that merges).

`antseed seller setup` previously asked for service capabilities as a raw `Capabilities JSON (optional):` blob, leaving sellers to guess the schema. The wizard now builds the JSON for them, one validated question at a time.

## Summary
- new guided flow in `capability-prompts.ts`: one question per capability field, inline validation via the shared `validateServiceCapabilityFields` (re-asks on invalid answers), empty answers mean "unknown" and are omitted, and the assembled JSON is echoed back at the end
- the question set depends on the service's protocol, classified from the upstream model / service id:
  - image models (gpt-image, dall-e, grok-imagine) → input modalities + supported request parameters only, with a note that `outputs: ["image"]` is announced automatically
  - text models → context window, max output tokens, inputs, outputs (hinted for multimodal chat models that also generate images), reasoning, tool use, structured output, supported parameters
- the image-model regex moved from the openai plugin into provider-core as `isImageModelId()` so the plugin's protocol advertising and the wizard's classification share one definition
- pasting a `{...}` JSON object at the opening prompt keeps the previous paste-the-JSON path, validated through the same parser

## Testing
- 8 scripted-prompt tests covering the full text flow, the short image flow, decline/skip-all, re-ask-until-valid, invalid modality/parameter rejection, and both JSON-paste paths
- full workspace typecheck; cli 298, provider-core 50, provider-openai 14 tests passing

## Changelog
`CHANGELOG.md` updated (Unreleased → Added); configuration guide updated to describe the guided flow.


kotevcode:
@claude

kotevcode:
Addressed in eb61f3f8: `contextWindow`/`maxOutputTokens` answers now go through `validateServiceCapabilityFields` like the other fields, so out-of-range values (> 1_000_000_000) are re-asked in the wizard instead of surfacing at announce time — guided flow and JSON-paste path now validate identically. Also added the upper-bound case to the re-ask test and aligned the test file's semicolon style with its siblings.

PR #811, opened 2026-08-11
feat(contracts): add seller position init faucet

kotevcode:
## Description

At cutover nobody holds transferable ANTS, so no pool has power and usage accounting ignores every settlement — no recognized-usage rewards can accrue for anyone. `AntseedPositionInit` is an unowned, immutable faucet that lets each existing legacy seller (agent binding + stake above the legacy minimum in the legacy USDC staking contract) create one small pre-funded ANTS position in their own agent pool via `stakeFor`, turning their accounting on from the first rewarded epoch.

Design points:
- All starter positions share one immutable `initEndEpoch` rather than a fixed lock duration. Position power is `amount * (endEpoch - e)`, so every starter position has identical power at every epoch regardless of claim time — a late claim never outweighs an early one.
- No admin, no sweep, no pause, no setters. Two natural off-switches: the funded pot running out (`InitDepleted`, refundable by topping up) and the end epoch being reached (`InitExpired`, permanent).
- One position per agent id, forever.

Also:
- `IAntseedSellerPools`: expose the existing `stakeActivationDelay` public getter so the faucet computes the same activation epoch as `stakeFor`.
- `DeployRecognizedUsage`: deploys the faucet (envs `POSITION_INIT_AMOUNT`, default 1 ANTS; `POSITION_INIT_END_EPOCH`, default effectiveEpoch + 104), whitelists it as an ANTS transfer sender, and extends the post-deploy checklist with funding and claim-window steps.

## Testing

- `test/AntseedPositionInit.t.sol`: 10 tests — position ownership/terms, power activation at the next epoch, late-vs-early equal power, expiry at the end epoch, one-init-per-agent, non-seller and below-min-stake rejection, depleted-pot top-up cycle, transfer-whitelist requirement, constructor validation.
- Full forge suite green.

Changelog deliberately not updated on this branch (entries added once before merge, per branch policy).

### DIFF
--- a/packages/contracts/interfaces/IAntseedSellerPools.sol
+++ b/packages/contracts/interfaces/IAntseedSellerPools.sol
@@ -110,6 +110,7 @@ interface IAntseedSellerPools {
 
     function antsToken() external view returns (IERC20);
     function currentEpoch() external view returns (uint256);
+    function stakeActivationDelay() external view returns (uint256);
     function agentIdForSeller(address seller) external view returns (uint256);
     function positionWithdrawableEpoch(uint256 positionId) external view returns (uint64);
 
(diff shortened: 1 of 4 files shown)


alexanderludwig:
Superseded by #968, which incorporates the position-init faucet into the corrected shared recognized-usage contract foundation. Closing this standalone PR so the branch stack remains: #968 as the base, with wash trading and KBF verification as independent sibling feature PRs.

PR #813, opened 2026-08-11
blog: Claude Code Pricing: When $17 Stops Being Cheap

Makatzrina:
New cost post. Works out where the $17 Claude Code subscription stops being the cheaper option, using published rates and live network data rather than estimates.

The spine: output tokens cost 5x input on every Claude model, but real agent traffic sends 6.32 input tokens per output token, so input ends up being 56% of the bill. That is the opposite of how most cost comparisons are framed.

## Sources verified

All figures pulled and checked on 2026-08-11.

- https://platform.claude.com/docs/en/about-claude/pricing - list rates for Opus 5 ($5/$25), Sonnet 5 ($2/$10 introductory through Aug 31, $3/$15 after), Haiku 4.5 ($1/$5), Fable 5 ($10/$50); cache read at 0.1x input; the note that Claude 4.7+ models emit ~30% more tokens for the same text
- https://claude.com/pricing - $17/month Pro on an annual subscription, includes Claude Code
- https://network.antseed.com/stats - 13,984,528 requests, 88.96B input tokens, 14.07B output tokens, 63 peers live. Average request shape of 6,361 input / 1,006 output derives from those totals
- Marketplace medians in the comparison table computed from the same feed, per model, across the providers quoting each service

Derived numbers (breakevens of 298, 746, 1,491, 149, and 545 with caching) are arithmetic on the two sources above and are shown with their inputs so a reader can redo them.

## Internal links added

- `/integrations/claude-code`
- `/vs/openrouter`
- `/docs/pricing`
- `https://network.antseed.com/stats`

Suggested follow-up, not done here: the existing OpenRouter comparison post could reasonably link back to this one. Left for a separate change rather than editing another post in this PR.

## Balance

The post keeps the figures that do not favour us: Sonnet 5's market median sits above Anthropic's introductory list price, and Anthropic's flat 10x cache discount beats the network's 7.7x median. There is a full section on when to just pay the $17. The boundary against reselling subscription credentials is stated.

## Please check before merging

- The network totals cover all traffic, not Claude Code specifically. The post says so explicitly, but confirm that framing is how we want to describe the endpoint.
- Sonnet 5 introductory pricing expires 31 Aug 2026. Two tables and one paragraph go stale that day. Worth a calendar reminder to update, or a follow-up post.
- Live marketplace medians move. The table is a dated snapshot and labelled as such; confirm we are comfortable publishing a point-in-time table that will drift.
- Not from a `blog-queue.yml` entry, so it does not have a reserved slot in the schedule.

Lint: `blog-lint.mjs` passes clean, 0 errors and 0 warnings. 1,367 words, 7 H2, 4 internal links.

### DIFF
--- a/apps/website/blog/2026-08-11-claude-code-pricing-breakeven.md
+++ b/apps/website/blog/2026-08-11-claude-code-pricing-breakeven.md
@@ -1,0 +1,115 @@
+---
+slug: claude-code-pricing-breakeven
+title: "Claude Code Pricing: When $17 Stops Being Cheap"
+authors: [antseed]
+tags: [pricing, Claude, cost, AI infrastructure, P2P, API pricing]
+description: Claude Code pricing breaks even near 298 requests a month. Here is the math, from Anthropic's published rates and 14 million real agent requests.
+keywords: [claude code pricing, claude code cost, claude code pro vs api, claude api pricing, claude opus 5 price, cost per request, prompt caching cost]
+image: /og-image.jpg
+date: 2026-08-11
+---
+
+Claude Code pricing comes in two shapes: a $17 per month Pro subscription, or per-token API billing where you pay for exactly what you use. The switchover point is sharper than most people expect. At Anthropic's published Opus 5 rates and a realistic agent workload, the subscription stops being the cheaper option at roughly **298 requests a month**. That is about ten a day.
+
+Below that, pay the $17 and stop reading. Above it, the arithmetic starts to matter, and it matters in a direction most cost comparisons get backwards.
+
+<!-- truncate -->
+
+## What does Claude Code actually cost?
+
+Claude Code is billed two ways. The Pro plan is $17 a month on an annual subscription and includes Claude Code alongside the chat product. The API route bills per million tokens, split into input and output, with a separate and much cheaper rate for cached input.
+
+Here are the published API rates, taken from [Anthropic's pricing documentation](https://platform.claude.com/docs/en/about-claude/pricing). Verified August 2026.
+
+| Model | Input / MTok | Output / MTok | Cache read / MTok |
+|---|---|---|---|
+| Claude Opus 5 | $5.00 | $25.00 | $0.50 |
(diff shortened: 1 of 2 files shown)


Makatzrina:
### Update: pricing reworked onto the $20 monthly price

The post was originally built on $17, which is the annual-billing rate. The number people recognise and search for is $20 monthly, so the title and the whole post now run on that, and the full ladder is in a table: Pro $20 monthly, $17 annual billed $200 up front, Max 5x $100, Max 20x $200.

Breakevens recomputed, with a $200 column added:

| Model | Cost per request | To reach $20 | To reach $200 |
|---|---|---|---|
| Opus 5 | $0.0570 | 351 | 3,512 |
| Sonnet 5 (intro) | $0.0228 | 878 | 8,779 |
| Haiku 4.5 | $0.0114 | 1,756 | 17,558 |
| Fable 5 | $0.1139 | 176 | 1,756 |

With a 90% cache hit rate: 641 against Pro, 6,412 against Max 20x. Chart redrawn on the $20 line with the x axis extended to 800 so both crossings fit.

### One number needs a human check before merge

**Max 20x at $200/month is not verified.** `claude.com/pricing` renders both Max tiers as "From $100" and I could not extract a distinct figure for the 20x tier from the page, the Max page, or the Claude Code support article. The $200 came from the author, not from a primary source. Everything else in the post traces to one.

Verified directly: Pro at $20 monthly and $17 annual billed $200 up front, and every API rate in the model tables.

Please confirm $200 against a live account before this goes out, or tell me and I will drop the Max rows.

### Staleness mitigation

Putting a specific price in the title bakes in a decay date, so the post now:

- cites the plans page as the source of the plan table, dated August 2026
- states outright that it is a snapshot and to check the live page before budgeting
- gives the method rather than just the answer, monthly fee divided by cost per request, so it still works after a repricing
- adds "check what your plan costs today" as step zero of the self-check list

The cost-per-request figures come from published API rates and measured token counts, and those do not move when a subscription tier is repriced.

PR #825, opened 2026-08-12
fix(node): encrypt p2p tcp transport and sign webrtc sdp

kotevcode:
## Description

P2P payload traffic on the TCP transport was plaintext. Adds `transport.tcp-enc.v1`: a wallet-signed ephemeral X25519 handshake (mutual auth, forward secrecy) with ChaCha20-Poly1305 framing, used automatically between peers that advertise it and failing closed once offered — no plaintext downgrade. Adds `transport.signed-sdp.v1`: WebRTC SDP messages are signed, binding the DTLS fingerprint to the peer identity so signaling MITM cannot substitute SDP. New `requireSecureTransport` node option rejects plaintext/unsigned transports entirely; default off for legacy interop.

Also activates the WebRTC transport, which had never run in production: `ConnectionManager.init()` was never called, so node-datachannel never loaded and every connection silently used TCP. Nodes now load it at startup, advertise `transport.webrtc.v1` when the stack works, and initiators use WebRTC only toward peers that advertise it (TCP fallback for mixed fleets). Inbound `hello` signaling is refused cleanly instead of crashing when the local stack is unavailable.

Covered by unit tests and loopback integration tests (wire-capture proxy asserts no plaintext, MITM ack forgery rejected, downgrade fails closed, full WebRTC loopback with signed SDP). Full node + e2e suites pass. CHANGELOG updated.


kotevcode:
@claude

PR #832, opened 2026-08-12
feat(protocol): count generated images in spending auth metadata (v3)

kotevcode:
SpendingAuth metadata v3 adds an output_images counter (top-level and per-service) and credits a flat 1290-output-token equivalent per generated image into the token counters, with an estimated prompt token count as input when the upstream reports no usage. Attribution only: image billing stays on the per-unit price and the equivalents never enter cost verification. First four ABI fields are unchanged for legacy aggregate decoders; the services array is version-gated. Includes channels migration 005 for the new per-service column.


kotevcode:
@claude

PR #834, opened 2026-08-13
feat(models): Add network model routing

kotevcode:
## Description

Adds a shared network service catalog for the CLI and desktop app, exposing discovered models through the OpenAI-compatible `/v1/models` API. Model-only requests resolve canonical aliases, select the highest-reputation compatible peer, and fall back through eligible peers while preserving explicit peer pins.

Rate-limited peers receive up to three attempts before routing falls back to the next peer. The original unrelated Telegram changes are not included.

## Release Notes

- Added network-wide model discovery through `/v1/models`
- Added canonical model aliases across CLI and desktop catalogs
- Added reputation-ranked model-only routing with peer fallback
- Added same-peer retries for HTTP 429 responses before fallback

## Types of Changes

## Checklist

Supersedes #833 after renaming its head branch.


kotevcode:
@claude

kotevcode:
@claude

kotevcode:
@claude

kotevcode:
## Code review — 10 confirmed correctness findings

Deep multi-agent review of this PR (8 finder angles, 26 raw candidates, every reported finding independently verified against the branch and `origin/main`; 1 candidate refuted and dropped). Ordered roughly by severity.

---

### 1. Coding-only offers leak into unrestricted requests
`apps/cli/src/proxy/routing.ts:122` (also 176–177)

`findAdvertisedServiceOffer` / `selectAdvertisedServiceByProtocol` fall back to a `-coding-only` offer when it is the peer's only canonical match — violating the invariant stated in the comment directly above (lines 102–103). `withRoutedModel` then rewrites `body.model` to the restricted serviceId.

**Failure:** client requests `fable-5` (general chat); a peer advertises only `fable-5-coding-only`. `canonicalModelKey` strips the suffix so the offer matches, `unrestricted.length > 0 ? unrestricted : offers` returns the restricted offer, and the seller applies coding-only restrictions (refusals/policy errors) to a non-coding conversation. That peer can be top-ranked even when unrestricted peers for the same model exist.

### 2. Desktop auto-mode syncs a coding-only representative serviceId
`apps/desktop/src/renderer/modules/routing/proxy-sync.ts:22`

The catalog entry's representative serviceId is the cheapest priced row, which can itself be a coding-only variant (`model-catalog.test.ts:120-121` asserts `entry.serviceId === 'fable-5-coding-only'` — display name cleaned, id not). `proxy-sync.ts` posts it as the buyer default route; the CLI substitutes it for the `antseed` alias and `routing.ts:117-119` treats it as an *explicit* coding-only request. Ordinary chats (including the Telegram bridge) get routed to a seller-side coding-only service under the clean "Claude Fable 5" label.

### 3. Pinned-peer path lost lenient filtering → local 502 instead of dispatch
`apps/cli/src/proxy/buyer-proxy.ts:2289`

The pinned path switched from unconditional `'lenient'` to `'strict'` whenever a model is present. A session-pinned seller whose metadata doesn't canonically list the requested model (stale DHT enrichment, newly added model, a tool's helper-model sub-requests like Claude Code's haiku calls) strict-misses → empty routing peers → local 502 "does not support this request" without ever contacting the seller. `routing.ts:29-31` still documents lenient as guaranteeing "a pinned peer still dispatches and surfaces the seller's real upstream error" — that mode is now unreachable for any request carrying a model.

### 4. Trust gate: hard cutoff at new default 60, unknown reputation ineligible, stored 0 silently migrated
`apps/desktop/src/renderer/modules/routing/select.ts:123`

Trust changed from a −50 scoring penalty (old default 0) to a hard eligibility gate defaulting to 60 that also rejects rows with *unknown* reputation, and `preferences.ts:84-87` migrates any un-versioned stored 0 — including a deliberate user choice — to 60.

**Failure:** fresh install or network where discover rows carry no reputation data (enrichment unavailable, RPC down, young sellers): every row is rejected, `chooseBestVprRoute` returns null with no relaxation pass, and since the projection fallback was removed, app-profile connect gets "No model route available yet", default-route sync no-ops, and failover returns null — where pre-PR these rows were merely penalized and still routed.

### 5. Legacy conversations coerced to `pinned`, disabling failover
`apps/desktop/src/main/chat/engine.ts:554` (also `conversation-store.ts:49`)

Conversations written before `routeMode` existed have `peerId` set but `routeMode` undefined; `routeMode: c.routeMode ?? 'pinned'` makes the renderer failover gate skip them. When the remembered seller goes offline, the chat 502s on every send forever instead of failing over. `peer-selection.ts:25-28` still documents the opposite contract ("threads written before this field existed... are treated as auto by callers") and no load-time migration stamps legacy records.

### 6. `reminder.ts` still keys the OpenRouter price map with legacy `normalizeModelKey`
`apps/desktop/src/renderer/modules/app/reminder.ts:109` (also 197, 446)

`openrouter-catalog.ts:55` now keys with the shared `canonicalModelKey` (`'fable5'`) but `reminder.ts` looks up with `normalizeModelKey` (`'claudefable5'`); the map travels verbatim over IPC. The PR migrated the sibling consumers (`measured-savings.ts`, `openrouter-baseline.ts`) but not this one. Retrospective savings misses all Claude-family usage, so the "worth $" figure is understated or falls below `MIN_AMMO_USD` and the deposit-reminder offer silently never shows.

### 7. Proxy-reuse probe fails against older published proxies
`apps/cli/src/cli/commands/buyer/start.ts:169`

`isCompatibleBuyerProxy` dropped all body-marker fallbacks and recognizes an existing proxy only by `x-antseed-*` response headers — which an older published proxy never sets on its locally-answered `/v1/models` 400 (`no_peer_pinned` sets only `content-type`; the telemetry headers are attached exclusively on peer-relayed responses). Version skew turns port reuse into "Failed to start proxy" / exit 1, where the old marker check reused the proxy.

### 8. Cold-start: catalog empty until live rediscovery
`apps/desktop/src/main/chat/engine.ts:303` + `apps/cli/src/proxy/buyer-proxy.ts:496`

The desktop catalog is now sourced exclusively from a live `/v1/models` fetch; the deleted `buyer.state.json` fallback's show-stale-peers policy ("filtering here leaves Discover empty on cold-start, e.g. laptop opened after overnight") is not re-established anywhere. Two layers reintroduce the regression: (1) fetch failure while the runtime is stopped/booting returns `lastServiceCatalogEntries` (initialized `[]`); (2) `parsePersistedPeers` discards every persisted peer older than the 2h `CARRY_FORWARD_TTL_MS`, so after overnight sleep `/v1/models` is empty until DHT rediscovery completes. Discover shows "Searching for services..." for the whole window despite a fully-populated on-disk peer cache.

### 9. Raw `onChainTrustScore` compared against the 0–100 gate
`apps/desktop/src/renderer/modules/routing/select.ts:44`

`modelReputationScore` falls back to raw `onChainTrustScore` — unbounded credited-USDC volume (desktop fixtures use values like `10_432`) — and compares it directly against `minTrustScore` without `scoreFromTrust` normalization, unlike node's own `normalizedModelReputationScore` for the identical fallback. Reachable with stale buyer-state files or `/v1/models` payloads from older CLIs lacking the first two fields: raw trust 100 (normalized ≈50) passes a gate of 60 and skews the ±50 scoring. Latent pre-PR, but the new hard gate + default 60 made the unit mismatch consequential.

### 10. Malformed percent-encoding in `/v1/models/:id` returns 502 instead of 404
`apps/cli/src/proxy/buyer-proxy.ts:1852`

`_handleNetworkModels` calls `decodeURIComponent` on the un-sanitized path segment; `GET /v1/models/gpt%zz` throws `URIError`, escapes to the server-level catch, and returns 502 `text/plain` "Proxy error: URI malformed" instead of the intended 404 `model_not_found` JSON (the decode inside the 404 message can throw the same way). Not a crash — wrong status/content-type for a client input error.

---

**Refuted during verification:** the `projection.ts:88-90` null-fallback change is deliberate (test-asserted: "auto mode does not bypass the minimum reputation gate") and every caller handles the null.

**Below the cap (cleanups, confirmed but outranked):** duplicate `CODING_ONLY` regex in `routing.ts:104`; uncached per-request `/v1/models` catalog rebuilds; per-call `buildNetworkServiceOffers` rebuilds in `routing.ts:113`.

kotevcode:
@claude review the last fix

Issue #856, opened 2026-08-15
Add CLI support for claiming ANTS rewards


kc-zero-lab:
Users who keep AntSeed running through CLI / external agents currently need to use VPR for claiming ANTS rewards. Since CLI remains the right path for VPS/Hermes/headless setups, add CLI support for claiming ANTS directly.

Context from Telegram:
- Some users still use `antseed` CLI as their provider path with old accounts.
- VPR currently makes claiming ANTS easier.
- Need parity so headless/CLI users don’t have to switch to VPR just to claim.

Acceptance criteria:
- CLI command for checking claimable ANTS.
- CLI command for claiming ANTS.
- Works with the same signer/account identity used by CLI payments.
- Clear output for no rewards, pending rewards, claimed amount, tx hash/errors.
- Docs/help text included.

PR #862, opened 2026-08-16
feat(cli): deposit flow, daemon auto-sweep, activity view

kotevcode:
## Description

Aligns the CLI payments interface with the desktop app's deposit and activity flows, and moves the auto-sweep engine into the buyer daemon so both surfaces share one signer.

- **`antseed buyer deposit`** is now the deposit flow: funding address + terminal QR (EIP-681), a browser-wallet checkout link (serves the pay page while watching), and live sweep progress until credited. Works through the running daemon or standalone with an ephemeral node. The old direct on-chain deposit moved to `--onchain <usdc>`; `antseed deposit` is a hidden alias.
- **Daemon auto-sweep**: `buyer start` now watches the hot wallet and sweeps incoming USDC gaslessly by default (`buyer.autoSweep`, default `true`), exposed on the control plane (`GET /_antseed/deposits/status`, `POST /_antseed/deposits/watch`). The desktop's `deposit-sweep.ts` becomes a thin shim that drives the daemon's watcher instead of running a second signer against the same wallet.
- **`antseed buyer activity`**: terminal counterpart of the VPR Activity view — tokens/spent/saved (vs a configurable retail-prices API), per-day spending chart (7/30/90d), active channels with channel IDs, close guidance, deposits balance, and claimable ANTS emissions. `--json` for scripts.
- **Locked-amount fixes (desktop + CLI)**: display the recoverable amount (`deposit − max(settled, cumulativeSigned)`) instead of the raw reserve; treat the chain as the authority on channel liveness (closed-on-chain rows are dropped); resolve channels behind delegated-seller facades via `channelsAddress()`; hide (but count) locally-tracked channels with no on-chain record. The desktop Activity header now sums the displayed rows.
- **`antseed payments`** no longer launches the retired portal — it prints the replacement commands. `buyer sweep` fails fast when every relayer declines.

CHANGELOG.md updated for all user-facing changes. Docs updated: CLI README, website CLI reference, payments/API guides, integrations pages, hermes-antseed skill.

#856 

## Testing

- CLI: full suite passes (`node --test`, 415+ tests), including new deposit-watcher unit tests and control-plane endpoint tests.
- Desktop: main-process suite (241) and renderer suite (297, incl. 12 updated/new locked-amount tests) pass; main + renderer typecheck clean.
- Smoke-tested: `payments`, `buyer deposit --no-watch` + QR, watch mode with web checkout (portal boots, URL served, Ctrl+C exits), `buyer activity` against a stub daemon and the no-daemon error path.
- **Configurable savings baseline**: the retail reference prices behind the Saved tiles (CLI + desktop) and Home strikethroughs are no longer hard-coded to OpenRouter — the source is any OpenRouter-schema-compatible models API set via `ANTSEED_COMPARABLE_PRICES_URL`; unset turns the baseline off.

- **Release baking + npm publish CI**: `ANTSEED_COMPARABLE_PRICES_URL` is baked into release artifacts at build time (`scripts/bake-comparable-prices-url.mjs` rewrites committed null-default modules; runtime env still overrides, empty value disables) — wired into desktop release jobs and a new manual `npm-publish.yml` workflow. The workflow publishes whatever versions were bumped (`pnpm -r publish` skips versions already on npm) and gates on `scripts/npm-publish-plan.mjs`, which prints the publish plan and fails on exact-pin dependents missing a bump or local versions behind the registry. Requires the `NPM_TOKEN` secret; the URL comes from the `ANTSEED_COMPARABLE_PRICES_URL` repo variable (already set).
- **Review fixes**: `buyer.autoSweep: false` in config is now honored (was dropped by the config merger), CLI activity reads channel input tokens from the daemon's `tokensDelivered` field (was always 0), and the desktop Active-channels row filter uses the recoverable-amount formula so fully-signed-away channels no longer linger.


kotevcode:
@claude review this pr

PR #876, opened 2026-08-16
fix(cli): Diagnose seller reachability

### LINKED ISSUE #848: Node announces "Seeding active" and passes health checks while completely undialable — no warning on empty publicAddress
## Problem

Reported by an external provider behind NAT: with `publicAddress` unset/empty, the seller starts, prints "Seeding active", and passes its own health checks — while being completely unreachable from the network. Nothing dials the announced address from outside, so the operator has no signal that no buyer can ever connect. This was their longest debugging session of the day.

Documentation gaps make it worse: `publicAddress` appears once in the config reference with no networking guide, and `/docs/protocol/transport` 404s.

## Expected

- Warn loudly at startup when the announced address is empty, private, or unroutable.
- Add a `seller doctor` (or extend `seller status`) that dials the announced address **from outside** — e.g. via a bootstrap/relay echo — and reports "your announced endpoint is not reachable" with NAT guidance.
- Write the networking/transport docs page (port forwarding, VPS/reverse-tunnel patterns) and fix the 404.

## Where

- `apps/cli/src/cli/commands/seller/start.ts` startup validation
- docs site transport page

kotevcode:
## Description

Warn sellers when their announced endpoint is missing or non-public, and add `antseed seller doctor` for address classification plus a bounded local TCP probe. Expand the canonical `/docs/transport/` guide with NAT, firewall, VPS, and reverse-tunnel setup guidance.

## Release Notes

- Seller startup now loudly warns about missing, private, loopback, link-local, multicast, and unroutable announced addresses.
- `antseed seller doctor` reports endpoint classification and local TCP connectivity, with an explicit reminder to verify NAT ingress externally.
- The existing `/docs/transport/` guide now includes practical seller networking and deployment guidance.

## Types of Changes

## Checklist

- [x] I have added the necessary documentation
- [x] I have added tests
- [x] CLI tests and typecheck pass
- [x] Website build passes

Fixes #848


kotevcode:
@claude review this add

kotevcode:
Addressed the review feedback:

1. **Duplicate host:port parsers** — consolidated into a shared `parseHostPort` helper (`apps/cli/src/config/public-address.ts`), promoted from the reachability version (the one that strips IPv6 brackets). `parseSellerEndpoint` and config validation's `parsePublicAddress` now both delegate to it; `parsePublicAddress` keeps its `MAX_PUBLIC_ADDRESS_LENGTH` guard on top. `extractHostFromAddress` in `proxy/request-utils.ts` was left as is — it extracts a host from an address *without* requiring a port, so it has genuinely different semantics.
2. **198.18.0.0/15 classification** — moved from `private` to `unroutable` alongside the TEST-NET ranges; `publiclyRoutable` is unchanged, only the label is now accurate. Added test cases for 198.18.0.1 and 198.19.255.255.
3. **Doctor probe skip** — `seller doctor` now runs the local TCP probe for any parsed endpoint (including private/loopback), so "is anything even listening" can be sanity-checked when the address classifies as non-public. `healthy` still requires a potentially-public address plus a successful probe, so the exit code semantics are unchanged.
4. **Hostname resolution** — left as a documented limitation per the review.

CLI test suite: 426 passed.

PR #895, opened 2026-08-18
feat(contracts): add standalone wash-trading enforcement

kotevcode:
## What

Makes wash-trading enforcement a standalone PR based directly on `main`.

This PR includes:

- The SP1-backed `AntseedWashTradingRegistry` and journal interfaces.
- Wash-only pair-aware points and historical seller-reward policies.
- Proof submission, BlockhashStore backfill, development submission, and volume-verification tooling.
- Idempotent deployment/configuration scripts for the wash-only registry and policies.
- Max-ratio materialization: overlapping claim IDs never add wash volume; only a seller's greatest individually proven wash ratio is retained.

## Isolation from contracts-update-all

- The PR has no commits from `contracts-update-all` or `refactor/contracts-domain-architecture` and is based on `main`.
- It modifies no Solidity contract that already exists on `main`.
- `AntseedEmissionsV2.sol`, `AntseedSellerRewardsPool.sol`, and `IAntseedSellerClaimPolicy.sol` remain byte-for-byte identical to `contracts-update-all` and `main`.
- The only path overlap with `contracts-update-all` is shared documentation/configuration: `CHANGELOG.md`, `packages/contracts/.env.example`, and `packages/contracts/README.md`.
- The prior combined work is preserved at `archive/pr895-original-2026-08-28` and `archive/pr895-pre-standalone-2026-08-28` for migration review.

## Migration safety

- Legacy one-sided `accrueSellerPoints()` and `accrueBuyerPoints()` behavior remains unchanged.
- Wash penalties apply only through the pair-aware points hook.
- Historical proportional reward claims require the proof backfill to complete and legacy emissions to be permanently paused.
- A proven seller snapshots the exact locked balance and can receive one proportional payout only while the pool still matches that snapshot; repeated or unexpected balance changes fail closed.
- During the replacement-emissions handoff, operators first confirm the replacement points policy, then pause legacy emissions and activate the historical reward policy while the protocol registry still references legacy emissions, and only afterward switch the registry.

## Validation

- `cd packages/contracts && forge test` — 608 passed, 0 failed.
- Proof-tool Node tests — 17 passed, 0 failed.
- Anvil-oriented lifecycle tests — 3 passed, 0 failed, 1 optional Base fork test skipped because `BASE_MAINNET_RPC_URL` was not set.
- Focused Solidity formatting and `git diff --check` pass.
- Archive port audit — all 34 final files exist in the preserved pre-standalone tree: 20 byte-identical and 14 intentionally adapted for `main`; no final file was omitted from the archive comparison.


kotevcode:
@alexanderludwig — heads-up from the review on AntSeed/loop-proof#2, since it lands in this contract.

Your evidence-bound claim IDs make `claim_id` no longer an economic identity, while this registry's only replay guard is `claimed[journal.claimId]` and wash volume accumulates:

```solidity
if (claimed[journal.claimId]) revert AlreadyClaimed();
...
r.washVolume += s.washVolume;   // accumulates
if (s.settledVolume > r.settledVolume) r.settledVolume = s.settledVolume;  // max
```

Two different evidence selections for the same seller/funder/cohort produce two distinct claim IDs, both pass the guard, and `washVolume` double-counts while `settledVolume` stays pinned — inflating `washRatioBps`. I produced two honest evidence sets for the same claim in one afternoon (1,437 vs 3,539 fundings), so this is reachable without an adversary.

Proposed fix (option 1 from the loop-proof review): add an `economic_id` to `WashJournal` — hash of seller/funder/cohort, no evidence hash — and have this registry dedupe on it:

```solidity
mapping(bytes32 => bool) public economicClaimed;
...
if (economicClaimed[journal.economicId]) revert AlreadyClaimed();
economicClaimed[journal.economicId] = true;
```

`claim_id` keeps evidence binding, genuinely disjoint cohorts still sum correctly, and the registry stays ~100 lines. Happy to implement across the journal, guests, and this contract if you agree with the shape.

alexanderludwig:
Standalone-main migration audit (August 28, 2026):

- Rebuilt `feat/wash-trading-registry` as one commit directly on current `main` (`c2e5af3dc`).
- PR now contains 39 wash-related files and no commits/files from `contracts-update-all` or `refactor/contracts-domain-architecture`.
- Compared against the preserved original wash PR delta: 31 files are byte-identical, 8 were intentionally adapted for current `main`, and 2 foundation-only artifacts were omitted (`packages/contracts/snapshots/usage-accounting.json` and `packages/contracts/test/AntseedLegacyRewardsPoolRegistry.t.sol`).
- Main-specific adaptation installs the wash points policy directly through `AntseedEmissionsV2.pointsPolicy`, preserves the cumulative reward entitlement protection, and fails closed for seller points if the wash registry is unavailable.

Preserved comparison branches:
- `archive/pr895-original-2026-08-28` at `1b584c6f8`
- `archive/pr895-pre-standalone-2026-08-28` at `fe7991bc6`

Validation completed locally:
- `forge build`
- Full `forge test`: 604 passed, 0 failed
- Wash/reward targeted Solidity suites
- Proof-tooling Node tests: 18 passed
- Production-shaped Anvil lifecycle: 2 passed, 1 optional Base-fork test skipped

kotevcode:
1. refactor(contracts): finalize wash proofs on last chunk and batch chunk authentication
     - finalizeSellerProof is gone. The seller record updates when the last chunk is authenticated.
     - New authenticateBlockReferenceChunks so several chunks go in one tx. The batch helper in test/mocks was a sign
       the contract's granularity was wrong, so it is deleted.
     - Weaker-or-equal proofs are rejected at staging time instead of after paying to authenticate every chunk.
     - Removed the dead checks (TooManyBlockReferences, the in-chunk ordering check, the redundant length check, the
       duplicate reference counter) and collapsed the five proof* views into stagedProof(proofId).
     - Journal ABI untouched, 9 tests green.
  2. refactor(contracts): merge wash proof submit scripts into one resumable submitter
     - The two Anvil-only submit scripts (396 lines, no way to hit Base) are replaced by
       scripts/submit-wash-trading-proofs.mjs. It handles development and production artifacts, can reuse a deployed
       registry, resumes safely, and targets Base with --allow-live plus SUBMITTER_PRIVATE_KEY.
     - Smoke-tested end to end on Anvil with a synthetic development artifact, including the resume path.

  Direction change for the registry

  The current design is one immutable period per registry with one max volume per seller. That does not match how we
  will use it. Emissions on 632 run in 7-day epochs from GENESIS = 1_775_728_461, and any policy that consumes the
  registry will ask about epochs, not blocks. So:

  - Key records by (seller, epoch). GENESIS and EPOCH_DURATION become constants equal to the gate's. No period
    immutables, one registry forever.
  - Journal commits epoch alongside periodStartBlock and periodEndBlock. The guest proves those two blocks are exactly
    the epoch's edges from the four boundary headers. Their hashes are already among the block references the chain
    authenticates against Chainlink, so the mapping is proven, not trusted.
  - Views: sellerRecord(seller, epoch) and provenWashVolume(seller, fromEpoch, toEpoch). Drop isProvenWashTrader, it is
    just volume != 0.
  - Prover slices claims per seller per epoch instead of one multi-month period. Since the README says every predicate
    parameter is still a calibration placeholder, re-proving is coming anyway, so the journal change is free now and
    expensive later.

  One consequence to design for on the 632 side: a proof for epoch N can only exist after N closes plus proving time,
  so enforcement is retroactive by at least one epoch. The natural hook is the claim policy withholding rewards for
  proven epochs, not the points hook zeroing live usage.

  Second simplification this enables

  Per-epoch proofs reference far fewer blocks than a multi-month one. If every proof fits in one transaction (roughly
  2,000 references ABI-encoded under Base's 128 KiB limit, the canary has 202), the journal can commit a single
  keccak256 over the block list and the whole staged-chunk flow goes away: verify proof, decode, check the list hash,
  check each hash against Chainlink, store if greater. About 120 lines, no partial state.

  I need the real block-reference counts per seller per epoch to confirm this. Please run the planner on the actual
  artifacts and send me the numbers.

  Tooling

  Under the historical requirement, the Chainlink store and the header-walk backfill are load-bearing and I am not
  fighting them. What I want cut:

  - Keep one anchor-discovery path in the planner, not both the storage-scanner catalog and the live search.
  - Executor: drop --in-flight pipelining, the checkpoint file, and the plan-digest approval ceremony. Chainlink
    verifies every header, so a bad plan can only waste gas, and the store itself is the checkpoint. Keep --execute and
    --allow-live.
  - Fold the header prefetch and cache library into the executor. encodeRpcBlockHeader already rebuilds headers from
    plain eth_getBlockByNumber, so fetch on demand with a flat memo.
  - Move the backfill scripts out of packages/contracts/scripts. They are a one-off ops tool, not protocol surface.

  Rough outcome: about 1,900 lines of tooling down to 600 to 700, and the pipeline goes from five steps to three.

  Verification gap

  The last commit says two production seller proofs were staged, authenticated against Chainlink, and finalized on a
  Base-fork Anvil with the real verifier. I could not find the artifacts, a prover request ID, a gas figure, or a build
  attestation anywhere, and the PR's own validation section lists the fork test as skipped. The only checked-in
  end-to-end test uses a mock verifier and a mock store.

  Please add to the PR: the two artifacts (or where they live), the planner output for them, and the gas and elapsed
  time from the fork run. Until then I am treating the end-to-end claim as unverified.

  Housekeeping

  - PR description still lists points and claim policies that are not in the diff. Please update it.
  - loop-proof main has the old multi-seller journal. Only feat/aip4-conserved-loop matches this contract. Merge that
    first so the two repos agree, then we do the epoch journal change on top.
  - README says the backfill must finish before deploying the registry. It does not, the registry only reads the store
    at authentication time.

PR #909, opened 2026-08-20
fix(buyer): keep payments running through RPC outages and report true payment errors

kotevcode:
## Summary

- A failed 1.5s chain-RPC probe at buyer startup silently disabled payments for the whole session, leaving seller 402s unrecoverable (the buyer never signed the required catch-up SpendingAuth) while the app reported Healthy. Payments now stay enabled: a background `RpcHealthMonitor` in `@antseed/node` retries with backoff, the negotiator skips best-effort on-chain reads while the RPC is down, and everything recovers automatically once it answers. `ANTSEED_ENABLE_SETTLEMENT=false` remains the explicit opt-out.
- `/_antseed/deposits/status` now reports why the deposit watcher is absent (`payments-disabled` / `no-deposit-relay` / `external-daemon`) plus live payments and RPC health via the new `AntseedNode.getPaymentsStatus()`; the desktop banner names the actual cause.
- A seller 402 while payments are not running on the buyer is converted into a buyer-fault error stating it is not a balance problem, instead of forwarding the seller's raw `payment_required` body asking the user to add credits. Desktop chat errors now display the authored message from proxy/negotiator error bodies instead of raw JSON or generic HTTP text (buyer-fault detection stays spoof-safe: only the proxy-scrubbed JSON code field is matched).

## Testing

- New tests: `RpcHealthMonitor` (probe, fallback URL, background recovery), negotiator balance-precheck skip while unreachable, 402→buyer-fault conversion, deposits-status reason reporting, chat stream-stop message extraction.
- Full suites pass: node (995), buyer-core (11), CLI proxy/start/watcher (134), desktop main (252). Root typecheck clean.

Updates `CHANGELOG.md` (user-facing behavior changes).


kotevcode:
@claude review this pr

PR #912, opened 2026-08-20
fix(desktop): trust-gate free model pricing and defaults

kotevcode:
## Summary

- Catalog entry prices, the **Free** badge, and "save up to X%" tags are now computed only from sellers the routing trust gate can actually select. Previously a rep-0 seller's $0 offer labeled a model "Free" while sends routed to a trusted paid seller and failed with 402 for unfunded users. If no seller of a model passes the gate, prices fall back to all sellers — matching what routing does in that case.
- First-run default model selection now walks a curated free-model priority list (DeepSeek Flash, MiniMax, Haiku, Qwen 235B, Nemotron 3 Super, Gemma, Mistral Large) and, past it, picks the free offer from the highest-trust seller — instead of catalog order, which defaulted new users onto a barely-trusted seller.
- The Recommended list's free ride-along now requires the $0 offer to come from a gate-passing seller (`hasEligibleFreeSeller`, stamped at catalog projection time), so low-trust free models stay in All Models without being endorsed.

## Testing

- `vitest run --dir src/renderer` — 325 tests pass (5 new)
- `tsc -p tsconfig.renderer.json --noEmit` — clean


kotevcode:
@claude review this pr

PR #922, opened 2026-08-21
feat(desktop): Telegram bot media support both ways

kotevcode:
## Description

The Telegram bridge was text-only: generated images were dropped ("Done (no text reply)"), and inbound photos/documents hit a generic "Only text messages are supported" error. Part of this feature existed before (image sending, removed in #831's follow-up 1224a4c2 as an unrelated change during the discovery refactor); this PR restores it and extends it to full two-way media support.

Outgoing: `/image <prompt>` generates via the best `openai-images` seller within the buyer routing policy and delivers the result as a Telegram photo (document fallback for oversized/non-inline formats), with a live "Generating…" draft. Assistant replies that contain image file blocks now send those images instead of collapsing to a text placeholder.

Incoming: photos and documents are downloaded (Bot API 20 MB limit enforced), persisted as conversation attachments, and passed through the same multimodal attachment pipeline as the app's paper-clip flow — images reach vision-capable models, documents get text extraction. Captions become the message text; voice/video/audio get an explicit "not supported yet" notice instead of a generic error.

Also: image-only sellers are excluded from the bot's text-model picker and default-route fallback (they remain reachable via `/image`), and `/stop` now cancels in-flight image generations too. `PiChatEngine.sendMessageStream` gained an `attachments` option and a new `generateImage()` entry point so non-renderer surfaces can use both flows.

## Release Notes

- The desktop Telegram bot now handles images in both directions: `/image <prompt>` generates pictures on the network and delivers them as photos, and image attachments in agent replies are sent instead of being dropped
- Photos and documents sent to the bot are read by the agent like in-app attachments (captions become the message text); voice notes get a clear "not supported yet" reply
- Image-only sellers no longer clutter the bot's /model picker, and /stop also cancels image generations

## Types of Changes

## Checklist


kotevcode:
@claude review

PR #924, opened 2026-08-22
fix(desktop): Bound model discovery phases so slow machines still get rows

kotevcode:
## Description

The desktop's model-discovery pipeline did unbounded network work (per-peer metering, seller-domain metadata, a ~400KB stats download every cycle) and on slow machines exceeded the UI's 12s cutoff on every refresh — the model list stayed at "No models discovered yet" / "Loading models…" forever while the runtime looked healthy, with no trace in exported logs.

## Changes

- Hard per-phase time budgets in the discover-rows handler (catalog 8s, metering 5s, enrichment 4s, network stats 8s) with graceful degradation — a slow phase is skipped for the cycle, not fatal. Budgets hold even if a fetch ignores its abort signal.
- Network-wide seller stats now come from the Antscan explorer API (~20KB) instead of the ~400KB aggregator snapshot, cached 60s with a 30s failure cooldown and aggregator fallback. New `explorerApiUrl` in chain config.
- UI cutoff raised 12s → 30s as a last-resort backstop above the ~26s worst case.
- Discovery failures, recoveries, and slow-cycle phase timings are logged to the system log so exported logs capture this failure mode.

CHANGELOG.md updated (user-facing desktop fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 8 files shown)


kotevcode:
@claude review this pr

kotevcode:
Addressed the review in 22b0abf2:

1. `getNetworkStats` cache, in-flight promise, and failure cooldown are now all keyed by the URL pair, so a chain switch mid-fetch can't serve the previous chain's stats or inherit its cooldown. Added a test covering a failed chain and a healthy chain queried back-to-back.
2. Updated the stale 12s comment in controller.ts.
3. Extracted `raceBudget` to its own module with tests: in-budget result, never-settling task, rejecting task, and late background completion.

PR #948, opened 2026-08-25
nix: build the CLI package with pnpm2nix-nzbr

wanderer:
## Why

The prevous in nix package I submitted has the problem that whenever the lockfile changes we need to bump the hash in the package definition. 

## What

This gets around that by replacing fetchPnpmDeps + pnpmConfigHook with mkPnpmPackage from [FliegendeWurst/pnpm2nix-nzbr](https://github.com/FliegendeWurst/pnpm2nix-nzbr). Dependencies are fetched per-package using the integrity hashes in pnpm-lock.yaml (seeded into the pnpm store via pnpm store add), so pnpm-lock.yaml changes no longer require updating a fixed-output hash in flake.nix. The  down side is that the package is slower to build but it was not too noticeable to me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,11 +122,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: DeterminateSystems/nix-installer-action@main
 
-      # First build fetches each dependency tarball individually; keep the
-      # store between runs so lockfile churn only fetches what changed.
-      - uses: DeterminateSystems/magic-nix-cache-action@main
+      # Pinned SHA — DeterminateSystems/nix-installer-action @main as of 2026-08-31
+      - uses: DeterminateSystems/nix-installer-action@33c9ab3ef95cd57c164d9d6eb1f9a46338538d41
+
+      # Pinned SHA — DeterminateSystems/magic-nix-cache-action @main as of 2026-08-31
+      - uses: DeterminateSystems/magic-nix-cache-action@55718cd9ebf1737a061487d5ada21abf31e99a98
 
       - name: Build antseed CLI
         run: nix build .#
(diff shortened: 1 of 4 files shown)


wanderer:
@kotevcode what wrong with the pr?

kotevcode:
Please add ci for the build

wanderer:
@kotevcode done

kotevcode:
Thanks for adding the Nix CI - the successful build and smoke tests address my main concern.
One remaining question: pnpm2nix-nzbr is a relatively small, lightly adopted project. I understand the benefit is per-package hashing from pnpm-lock.yaml, avoiding manual fetchPnpmDeps hash updates.
Did you consider either keeping the standard nixpkgs approach or vendoring the small required Nix implementation into this repository? I’m not necessarily blocking the PR on this, but I’d like to understand why the additional external build dependency is preferable to the maintenance cost and supply-chain surface.
Also, could the two Determinate Systems actions be pinned to commit SHAs instead of mutable @main references?

wanderer:
> Did you consider either keeping the standard nixpkgs approach or vendoring the small required Nix implementation into this repository? 

Yes the standard is nixpkgs approach would be ideal for nix users, but it would require bumping this hash https://github.com/AntSeed/antseed/blob/main/flake.nix#L146 every time the deps change. The CI will now catch that and build failure will show the new hash so non-nix users would never have to touch nix but I thought this might be an inconvenience. I opened a new PR that bumps the hash and adds CI here #971 

Also on pnpm2nix-nzbr, it is a fork https://github.com/nix-community/pnpm2nix, i have just been using FliegendeWurst's fork since its maintained. The big reason that these are not more widely adopted is that they use [IFD logic](https://nixcademy.com/posts/what-is-ifd-ups-and-downs/) which is considered slow but there are lots work to speed it up.

kotevcode:
closing in favor of #971

PR #954, opened 2026-08-26
feat(desktop): Connect Claude Desktop via third-party gateway

kotevcode:
## Summary

Adds **Claude Desktop** to the desktop app's Connected Apps: connecting it switches Anthropic's Claude app into its native third-party inference mode, pointed at a new loopback gateway inside AntSeed Desktop. Chats in Claude then route through the buyer proxy to the network — no API key, no MITM proxy — and the user's normal Claude login, chats, and MCP config stay untouched in the separate 1p profile.

## How it works

Claude Desktop supports a third-party gateway mode: `deploymentMode: "3p"` in `claude_desktop_config.json` makes it boot against a separate `Claude-3p` profile directory whose applied `configLibrary` entry names an Anthropic-Messages-compatible gateway. The connect flow writes that profile and restarts Claude; disconnect restores it.

- **Claude gateway** (`apps/desktop/src/main/connected-apps/claude-desktop-gateway.ts`): loopback `node:http` server (default port 8380, `ANTSEED_CLAUDE_GATEWAY_PORT` override). Serves `GET /v1/models` in Anthropic's native catalog shape — Claude's picker only accepts its own model ids grouped by family, so entries are advertised behind Claude's ids with real names as display names: **AntSeed Auto** first (follows the route selected in VPR, live), then the top of the desktop's curated model picker. Forwards `/v1/messages` and `/v1/messages/count_tokens` to the buyer proxy with the model rewritten per slot, strips Claude's placeholder credential, rejects non-loopback Hosts and any request carrying an Origin, streams SSE through untouched.
- **`claude-desktop` config-patch format** (`system-proxy/config-patch.ts`): flips both deployment modes to `3p` and writes the `Claude-3p/configLibrary` gateway profile. Removal has an ownership guard — an enterprise/MDM-provisioned third-party profile is never unwound. On Windows, every known install layout is probed: classic `%APPDATA%\Claude`, MSIX/Store installs under `%LOCALAPPDATA%\Packages\Claude_*\LocalCache\Roaming\Claude`, and the `%LOCALAPPDATA%` / "Claude Nest" variants; apply writes the first existing root, remove unwinds every root it owns.
- **Profile entry** (`connected-apps/defaults.ts`): `open-tool` action opens Claude after connect (restarting it first when it was already running so it re-reads the profile). The Apps row, brand icon, and restart plumbing come from the existing Connected Apps surfaces — no renderer or IPC changes.
- **Gateway lifecycle** (`system-proxy/runtime.ts`): started before the patch applies so a gateway failure fails the connect; stopped on disconnect and runtime stop.
- **Chat attribution** (`apps/cli/src/proxy/conversation-identity.ts`): Claude Desktop titles each new chat from a session id of its own ("You are coming up with a succinct title…"); that prefix is now recognized as title housekeeping so it no longer creates a phantom conversation row per new Claude chat.

## Testing

- 59 desktop main-process tests (`node --test`), including new coverage for the gateway (catalog slots, model rewrite, SSE forwarding, loopback/Origin rejection, 502 fallback), the config patch round-trip with ownership guard, and Windows target resolution (candidate order, MSIX discovery).
- 31 CLI conversation-identity tests including the new title-request case; full renderer suite (333) green.
- Verified end-to-end on macOS against the real Claude Desktop app: connect rewrites the profile, Claude relaunches, chats route and bill through the network, and the desktop's route selection applies live.

## User-facing changes

CHANGELOG updated (desktop: Claude Desktop connected app).


kotevcode:
@claude review this

kotevcode:
Addressed the review:

- **Cursor card anchor** — intentional: the Apps list order was deliberately changed to Codex → Claude → Cursor → Hermes → Droid in a later commit on this branch; the anchor follows the Claude row by design.
- **Dead Windows paths in defaults.ts** — fixed in d70d0c73: the claude-desktop entry now carries plain posix paths with a comment pointing at `claudeDesktopPatchTargets`, which owns all Windows candidate resolution (classic/MSIX/Nest layouts).
- **Vestigial `providerKey`** — dropped from the claude-desktop format entirely (type, defaults entry, parser branch). The CLI-side profile parser now treats `providerKey` as optional since it never reads it — the desktop applies patches; the CLI only passes the payload through.

Desktop main 305/305 and CLI system-proxy 50/50 pass after the changes.

PR #955, opened 2026-08-26
fix(cli): fail over to a sibling peer when a pinned peer rejects

alexanderludwig:
Pinned routing treated the pin as an exclusive lock and surfaced the pinned peer's failure verbatim. A 403/401 was also counted as a routing success because it was outside the retryable set, so a forbidding peer kept healthy stats and was never bypassed.

Treat the pin as a preference: on a failover-eligible result (transport error, retryable 5xx/429/408, or auth rejection 401/403) try the next discovered peer that matches the service and passes buyer routing policy, only surfacing the error if every candidate fails. Record these statuses as routing failures, guarding against double-counting the transport-error failures _dispatchToPeer already records.

Adds regression tests for 403 failover and the no-sibling path, plus a CHANGELOG entry.

## What

Brief description of the change.

## Why

Why is this change needed?

## Testing

How was this tested? What tests were added or updated?

## Checklist

- [ ] `pnpm run build` passes
- [ ] `pnpm run test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,7 @@ This project uses selective package publishing. Each release entry lists the pub
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


alexanderludwig:
@claude review

PR #961, opened 2026-08-27
fix(protocol): Clarify peer failure responses

kotevcode:
## Description

Moves seller-failure guidance fully into the shared buyer protocol response so CLI/API and desktop consumers receive the same structured error, including pinned-route context, original seller message, and HTTP status. Non-success streaming responses are buffered before client callbacks, preventing empty or partially forwarded errors. Desktop preserves the protocol-authored message without adding generic stream text, and pinned conversations show retryable peer failures immediately instead of silently retrying the same peer.

## Release Notes

- Peer failures now return consistent guidance through the CLI/API and desktop, preserving the original seller response and status.
- Pinned-peer errors are displayed immediately, while Auto routing can continue failing over to another peer.

## Types of Changes

## Checklist

- [x] I have added the necessary documentation
- [x] I have added tests
- [x] Relevant builds and unit tests pass locally


kotevcode:
@claude

kotevcode:
Addressed review observation 2: transformed pre-stream seller errors now adapt the already-normalized peer response instead of the raw response, preserving AntSeed peer guidance across protocol translation. Added regression coverage for an Anthropic streaming request routed to an OpenAI Responses peer that immediately returns 503. CLI build and all 127 buyer-proxy tests pass.

kotevcode:
@claude review that last error message commit

kotevcode:
Addressed the latest review. This was an accidental null-safety bug, not intentional behavior: a pinned send failure can omit `stopReason`, so the non-optional access threw and the outer catch scheduled a same-peer retry.

Fixed the access to use optional chaining, added regression coverage for `{ ok: false, error: "Conversation not found" }` without a `stopReason`, and clarified the pinned-peer behavior in `CHANGELOG.md`.

Validation: `peer-routing.test.ts` passes all 41 tests; `git diff --check` passes. Commit: `7fd8f451`.

PR #980, opened 2026-09-01
blog: when does prompt caching actually save money

Makatzrina:
## What this adds

One blog post: **[When Does Prompt Caching Actually Save Money?](apps/website/blog/2026-09-01-when-does-prompt-caching-save-money.mdx)** — 1,245 words, six question-shaped H2s, FAQPage schema for answer engines.

Plus its cover SVG and the one-line slug registration that stops the blog card falling back to the generic placeholder.

## The angle

Every guide on prompt caching quotes "up to 90% savings" and almost none do the arithmetic on the **write premium**, which is what actually decides whether caching saves anything.

The post works the break-even for all three pricing shapes:

| Write premium | Ahead from |
|---|---|
| None | the 1st reuse |
| 1.25x (5-minute) | the 2nd use |
| 2x (1-hour) | the 3rd use |

Then a worked example at published Claude Opus 5 rates where input cost falls 82% but the bill only falls 67%, because output is never cached and is priced five times higher. That gap is the honest correction to the headline number.

It also names the trap: a one-shot workload that never re-reads its prefix **doubles** its input cost on a 2x write premium, and the long cache window looks like the safer option while carrying the higher premium.

## Sourcing

The post makes **no claims about AntSeed**. Every figure comes from OpenAI's and Anthropic's own pricing and caching documentation, fetched and checked 1 September 2026, so nothing here needs internal verification.

Target keyword: `when does prompt caching save money`. The head term `prompt caching` is held by established domains and isn't reachable from current authority; the question form is, and it's the shape answer engines quote.

## Verification

- All 3 internal links statically checked against real built routes — the site uses `onBrokenLinks: 'throw'`
- Both external sources fetched and every quoted figure verified against the live vendor pages
- Arithmetic in both tables recomputed independently

**Not run: a full `docusaurus build`.** Dependency install did not complete in the working environment, and `ci.yml` builds tiers 0–3 but never touches `apps/website`, so the website build is not exercised on PRs either. The only code change is one string added to a `Set<string>`. Worth a local `pnpm --filter @antseed/website build` before merge.

## Changelog

No user-facing package, CLI, desktop, protocol, or configuration change. Website content only, so `CHANGELOG.md` is not updated.

### DIFF
--- a/apps/website/blog/2026-09-01-when-does-prompt-caching-save-money.mdx
+++ b/apps/website/blog/2026-09-01-when-does-prompt-caching-save-money.mdx
@@ -1,0 +1,146 @@
+---
+slug: when-does-prompt-caching-save-money
+title: "When Does Prompt Caching Actually Save Money?"
+authors: [antseed]
+tags: [prompt caching, cost, API pricing, LLM API, developer tools]
+description: "Prompt caching is sold as a 90% discount. The real number depends on a write premium most guides skip. Here is the break-even math, with the vendors' own rates."
+keywords: [when does prompt caching save money, prompt caching, prompt caching cost, cached input tokens, prompt caching break even, llm cost optimization]
+image: /og-image.jpg
+date: 2026-09-01
+---
+
+import Head from '@docusaurus/Head';
+
+export const cachingFaqLd = {
+  '@context': 'https://schema.org',
+  '@type': 'FAQPage',
+  mainEntity: [
+    {
+      '@type': 'Question',
+      name: 'When does prompt caching save money?',
+      acceptedAnswer: {
+        '@type': 'Answer',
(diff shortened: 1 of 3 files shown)


Makatzrina:
## Build verified

Ran a full production `docusaurus build` against the repo-pinned pnpm 9.15.4 tree. This closes the caveat in the description.

```
✔ Server: Compiled successfully
✔ Client: Compiled successfully
[SUCCESS] Generated static files in "build".
```

Confirmed in the output:

- `build/blog/when-does-prompt-caching-save-money/index.html` renders
- `FAQPage` JSON-LD present in the emitted HTML
- Cover SVG copied to `build/img/blog/covers/`
- `onBrokenLinks: 'throw'` passed — all three internal links resolve

One warning, **pre-existing and unrelated to this PR**: a broken anchor on `/docs/guides/using-the-api/` pointing at `/docs/guides/agents/#define-your-internet-accessible-antseed-endpoint`. That heading no longer exists on the agents page. Worth a separate fix.

Note for whoever picks this up: `ci.yml` builds tiers 0–3 but never touches `apps/website`, so the website build isn't exercised on PRs. Given the blog is about to receive regular commits, adding a website build step to CI would catch this class of problem automatically.

PR #1001, opened 2026-09-11
feat(attribution): carry the GA session into the installer and report app milestones

lstern-polygon:
## Summary

Attribute desktop installs to the website session and Google Ads campaign that produced them, and report what people actually do after installing — so campaigns can be optimised toward users who open the app and use it, not toward downloads.

Today the join between website and app ends at the installer. The PostHog → GA4 webhook forwards `app_first_opened` under the app's install id, which GA4 cannot tie to any session, so every app event lands on source `(not set)` and Google Ads has never counted one (verified against GA4 property 548266550: 122 `app_first_opened` events since 28 Aug, 100% unattributed).

## How it works

1. **Proxy stamps the installer** (`apps/download-proxy/src/attribution.ts`). The website already appends the visitor's GA client id and session id to download links. The proxy now encodes them into a short HMAC-signed token and serves the file as `AntSeed-VPR-Setup-0.2.38.a-<token>.exe`. No storage: the token is the record. Tokens are filename- and URL-safe, expire after 30 days, and only prove origin (holding one lets you attribute events to that download session, nothing else).
2. **Installer records its own name** (`apps/desktop/build/installer.nsh`, new `nsis.include`). NSIS writes `$EXEFILE` to `installer-name.txt` beside the app and sends a fire-and-forget `GET /i?f=<name>` beacon (`installer_started`) via the `curl.exe` that ships with Windows 10 1803+. The signed bytes are untouched, so Authenticode is unaffected. AppImage needs nothing: the path is in `$APPIMAGE`. macOS has no channel (the .dmg name does not survive the drag to /Applications) and reports unattributed.
3. **App reports milestones** (`apps/desktop/src/main/telemetry/attribution.ts`, hooked in `telemetry.ts`). Read the token once, then `POST /app-events` with `{token, install_id, events}`:

| Event | When | Once per install |
|---|---|---|
| `app_first_opened` | first launch | yes |
| `app_onboarded` | setup completed **and** network joined, either order | yes |
| `app_activated` | first of: tool connected, first chat, deposit, routing mode change, API config copied, plugin installed — with `kind` | yes |
| `tool_connected` | `app_connect`, with the app from the existing fixed taxonomy | yes |
| `first_chat_started` | first chat, with `service_category` | yes |
| `deposit_completed` | every deposit, with `amount_bucket`, `is_first_deposit`, `days_since_first_open` | no |

4. **Proxy forwards to GA4** (`apps/download-proxy/src/app-events.ts`) under the original client id and session id. Fixed event catalog, param allowlist, 8 KB / 10 events per request. Events without a valid token still count under the app's random install id, flagged `attributed=0`.

Reports follow the same kill switches as PostHog telemetry (`TELEMETRY_ENABLED`, dev builds) and the in-app opt-out, and are off entirely unless an endpoint is baked (`BAKED_ATTRIBUTION_ENDPOINT`, set by `scripts/bake-desktop-telemetry-config.mjs`, default `https://download.antseed.com` for release builds; `ATTRIBUTION_ENDPOINT` overrides at runtime). What leaves the device: the token, a random per-install UUID, and the same coarse buckets PostHog already gets. No wallet address.

## Rollout

- `wrangler secret put ATTRIBUTION_SECRET` on the proxy (any long random string), then deploy. Unset = no stamping, no verification, everything else unchanged.
- In GA4, mark `app_activated`, `app_first_opened`, `first_chat_started` and `deposit_completed` as key events; the existing "app_first_opened" conversion action in Google Ads starts counting on its own, and `app_activated` needs one new imported conversion action, set as primary. `download_completed` should move to secondary.
- Retire the PostHog "HTTP Webhook" destination (its GA4 API secret is also in plain text there).
- First Windows release after merge produces stamped installers; opens from earlier installs keep reporting under install ids only.

## Not in this PR

- macOS attribution (needs a different channel; probabilistic proxy-side matching is an option).
- A "tunnel opened" milestone: the app has no such user action today, only the tunnels view being opened.

## Testing

- `pnpm --filter @antseed/download-proxy typecheck && test`: 37 tests pass (5 new for tokens, 7 for the endpoints).
- Desktop: `attribution.test.ts` (stamp parsing per platform, reporter, integration through `createTelemetryService`: once-per-install semantics across launches, onboarded ordering, opt-out, unconfigured endpoint) plus the existing 55 telemetry tests, 66 pass, run twice. Compiled in isolation because the monorepo packages were not built locally; the full `pnpm -C apps/desktop test` should be run in CI.
- `installer.nsh` is not compiled locally (no `makensis` on this machine) — please check the Windows release build output.
- Changelog updated (`Unreleased › Added`).


lstern-polygon:
Re-opened from the leonstern93-vibier account; superseded.

PR #1002, opened 2026-09-11
feat(attribution): carry the GA session into the installer and report app milestones

leonstern93-vibier:
## Summary

Attribute desktop installs to the website session and Google Ads campaign that produced them, and report what people actually do after installing — so campaigns can be optimised toward users who open the app and use it, not toward downloads.

Today the join between website and app ends at the installer. The PostHog → GA4 webhook forwards `app_first_opened` under the app's install id, which GA4 cannot tie to any session, so every app event lands on source `(not set)` and Google Ads has never counted one (verified against GA4 property 548266550: 122 `app_first_opened` events since 28 Aug, 100% unattributed).

## How it works

1. **Proxy stamps the installer** (`apps/download-proxy/src/attribution.ts`). The website already appends the visitor's GA client id and session id to download links. The proxy now encodes them into a short HMAC-signed token and serves the file as `AntSeed-VPR-Setup-0.2.38.a-<token>.exe`. No storage: the token is the record. Tokens are filename- and URL-safe, expire after 30 days, and only prove origin (holding one lets you attribute events to that download session, nothing else).
2. **Installer records its own name** (`apps/desktop/build/installer.nsh`, new `nsis.include`). NSIS writes `$EXEFILE` to `installer-name.txt` beside the app and sends a fire-and-forget `GET /i?f=<name>` beacon (`installer_started`) via the `curl.exe` that ships with Windows 10 1803+. The signed bytes are untouched, so Authenticode is unaffected. AppImage needs nothing: the path is in `$APPIMAGE`. macOS has no channel (the .dmg name does not survive the drag to /Applications) and reports unattributed.
3. **App reports milestones** (`apps/desktop/src/main/telemetry/attribution.ts`, hooked in `telemetry.ts`). Read the token once, then `POST /app-events` with `{token, install_id, events}`:

| Event | When | Once per install |
|---|---|---|
| `app_first_opened` | first launch | yes |
| `app_onboarded` | setup completed **and** network joined, either order | yes |
| `app_activated` | first of: tool connected, first chat, deposit, routing mode change, API config copied, plugin installed — with `kind` | yes |
| `tool_connected` | `app_connect`, with the app from the existing fixed taxonomy | yes |
| `first_chat_started` | first chat, with `service_category` | yes |
| `deposit_completed` | every deposit, with `amount_bucket`, `is_first_deposit`, `days_since_first_open` | no |

4. **Proxy forwards to GA4** (`apps/download-proxy/src/app-events.ts`) under the original client id and session id. Fixed event catalog, param allowlist, 8 KB / 10 events per request.
5. **Server-side match for installs without a stamp** (`apps/download-proxy/src/match.ts`) — macOS, renamed files. At `download_completed` the proxy stores a 48-hour Workers KV record keyed by platform and an HMAC of the client IP (raw IP never stored) with the GA ids, arch and ref. A milestone arriving without a token is matched to a record with the same platform and IP hash, accepted only when exactly one download fits (arch breaks ties; ambiguity = no match). The matched install id is remembered for 90 days; the download record is consumed so no second install can claim it.

Every forwarded milestone carries `attribution_method`:

| Method | How | Trust |
|---|---|---|
| `token` | app read the stamp from its installer filename (Windows, AppImage) | exact |
| `match` | proxy matched platform + IP hash to one recent download (macOS, renamed files) | probabilistic; degrades on shared IPs (CGNAT, offices, Private Relay) |
| `none` | neither; delivered under the app's random install id, `attributed=0` | count only |

An affiliate code on the download URL (`?ref=partner_42`) travels in the token and in the match record and is forwarded as `ref` on every milestone, so affiliate credit can be exact on Windows/Linux and probabilistic on macOS, and payouts can be limited by method.

Reports follow the same kill switches as PostHog telemetry (`TELEMETRY_ENABLED`, dev builds) and the in-app opt-out, and are off entirely unless an endpoint is baked (`BAKED_ATTRIBUTION_ENDPOINT`, set by `scripts/bake-desktop-telemetry-config.mjs`, default `https://download.antseed.com` for release builds; `ATTRIBUTION_ENDPOINT` overrides at runtime). What leaves the device: the token, a random per-install UUID, and the same coarse buckets PostHog already gets. No wallet address.

6. **Source signals that survive a blocked GA** (`apps/download-proxy/src/referrer.ts`, `apps/website/src/lib/analytics.ts`). 84% of unpaid installs — the group that produces most deposits — arrive with no GA session (ad blockers), so GA4 cannot say where they came from. Two additions: every download event now carries `referrer_host`, the public site the link was clicked on, from the `Referer` header, reduced to an allowlisted host (GitHub, X, Reddit, HN, Discord, the AI assistants, search engines, npm…) with anything private reported as `other` and no path or query ever read; and the website remembers the landing page's `utm_*` tags for the tab session and carries them onto proxy download links, which the proxy records as `link_source` / `link_medium` / `link_campaign`. Both are read server-side.

## Rollout

- `wrangler secret put ATTRIBUTION_SECRET` on the proxy (any long random string), then deploy. Unset = no stamping, no verification, everything else unchanged.
- `wrangler kv namespace create ATTRIBUTION_KV` and add the binding to `wrangler.toml` (commented block included). Without it, only the macOS / renamed-file match path is off. Storing an IP hash for 48 h is a privacy-policy line — please confirm you're fine with it.
- In GA4, mark `app_activated`, `app_first_opened`, `first_chat_started` and `deposit_completed` as key events; the existing "app_first_opened" conversion action in Google Ads starts counting on its own, and `app_activated` needs one new imported conversion action, set as primary. `download_completed` should move to secondary.
- Retire the PostHog "HTTP Webhook" destination (its GA4 API secret is also in plain text there).
- First Windows release after merge produces stamped installers; opens from earlier installs keep reporting under install ids only.

## Reviewer notes

- `telemetry-state.json` gains three fields (`attributionInstallId`, `attributionToken`, `attributionSent`); missing fields normalise to null / empty, so existing installs load unchanged and the schema version stays 1.
- Milestone marks are persisted by the calling method's own `save()`; the network call is detached and never touches disk, so no write outlives the method (the existing test suite's temp-dir cleanup relies on this).
- `build/installer.nsh` could not be compiled on the authoring machine (no `makensis`); it uses only `FileOpen`/`FileWrite`, the bundled `nsExec` plugin, and Windows' built-in `curl.exe`. The first Windows release build is the real check — please look for `installer-name.txt` beside the installed exe and an `installer_started` line in the proxy logs.
- Desktop tests were run by compiling `src/main/telemetry` in isolation (workspace packages were not built locally); `pnpm -C apps/desktop test` should run in CI.
- GA4 custom dimensions for the new params are already registered on the property (`kind`, `app`, `install_source`, `amount_bucket`, `is_first_deposit`, `days_since_first_open`, `service_category`, `attributed`, `referrer_host`, `link_source`, `link_medium`, `link_campaign`).
- The website change could not be type-checked locally (website deps not installed here); it is small and self-contained in `analytics.ts` plus one call in `Root.tsx`.

## Not in this PR

- A "tunnel opened" milestone: the app has no such user action today, only the tunnels view being opened.
- Matching on anything beyond platform + IP hash + arch (for example browser UA family). Can be added to the record if ambiguity turns out to be common.

## Testing

- `pnpm --filter @antseed/download-proxy typecheck && test`: 53 tests pass (6 new for referrer allowlisting and utm parsing) (6 for tokens incl. ref, 9 for the endpoints incl. match and method tagging, 7 for the matcher: single match, consumed record, remembered install, arch tie-break, ambiguity refused, window, no raw IP stored).
- Desktop: `attribution.test.ts` (stamp parsing per platform, reporter, integration through `createTelemetryService`: once-per-install semantics across launches, onboarded ordering, opt-out, unconfigured endpoint) plus the existing 55 telemetry tests, 66 pass, run twice. Compiled in isolation because the monorepo packages were not built locally; the full `pnpm -C apps/desktop test` should be run in CI.
- `installer.nsh` is not compiled locally (no `makensis` on this machine) — please check the Windows release build output.
- Changelog updated (`Unreleased › Added`).


leonstern93-vibier:
@kotevcode ready for review when you have a moment. Three commits: the installer stamp + app milestones, the server-side match for macOS you asked about (with attribution_method on every event), and referrer-host + carried utm tags on downloads. Reviewer notes in the description cover the state-file fields, the untested NSIS include, and the KV namespace.

